PR #31, opened 2016-01-04
Fixed #30

### LINKED ISSUE #30: options.className in constructor makes injection fail
example 

``` javascript
export class MyClass extends AnotherClass implements IMyClass {
    constructor( 
      IShouldbeInjected: IShouldbeInjected ) {
        var options: any = {}
        options.el = '#mydiv';
        options.className = "row";
        super(options);
        this.childView = ChildView;
        this.collection = IShouldbeInjected;
    }
}
```

comes from kernel.ts _getConstructorArguments at line 96 in
if ('function' === typeof Map &&
            fnStr.indexOf("class") !== -1 &&
            fnStr.indexOf("constructor") === -1) {

```
        result = null;
    }
```

you probably have to replace fnStr.indexOf("class") with fnStr.indexOf("class ")

jamesadarich:
Added checks for the keywords "class" and "constructor" instead of doing string.indexOf to allow properties containing the words "class" or "constructor" in the body of the constructor to not trigger the check.

### DIFF
--- a/source/kernel.ts
+++ b/source/kernel.ts
@@ -94,8 +94,8 @@ class Kernel implements IKernel {
     // If using ES6 classes and there is no constructor
     // there is no need to parser constructor args
     if('function' === typeof Map &&
-       fnStr.indexOf("class") !== -1 &&
-       fnStr.indexOf("constructor") === -1) {
+       /\b(class)\b/.test(fnStr) &&
+       /\b(constructor)\b/.test(fnStr) === false) {
 
       result = null;
     }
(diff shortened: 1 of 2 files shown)


remojansen:
@jamesrichford thanks a lot for your contribution :)

jamesadarich:
No problem! Invesrsify has made my life working with Marionette much easier. Let me know if you need any help with stuff would be glad to lend a hand :)

remojansen:
@jamesrichford sure, if you want to keep an eye on issues and send pull request your are more than welcome. I'm planing a lot of things for version 2.0 but I'm going to be busy with work for a few months :( Any kind of help is great :)

Issue #18, opened 2015-07-09
Typescript Definition File Not An External Module


pixelshaded:
```
declare module inversify {
```

should be  

```
declare module 'inversify' {
```

So that the following will compile

```
import inversify = require('inversify');
```

Right now its an ambient internal module and needs to be an ambient external module.

See 'Working with Other JavaScript Libraries' at http://www.typescriptlang.org/Handbook#modules

remojansen:
Hi @pixelshaded this issue should be fixed now by commits f3e7859b9301f9ef7f73c5abdf79c6b451c08eaf and pull request https://github.com/borisyankov/DefinitelyTyped/pull/5698 

It would be much appreciated if you can conform that everything works as you expected. 

Thanks!

remojansen:
I'm going to close this issue to ensure that we only have one unique issue to refer to this problem. You can follow the updates at #46

PR #32, opened 2016-01-09
Fixes I found while working with invserify and VSC and Visual Studio 2015

tenowg:
I am issuing the PR to bring these issues to light and offer a potential fix. These fixes do make this work with VSC and Visual Studio Code, I have not tested, nor do I know what implications this will have on other environments.

Please read commit comments for my reasons for these changes.

Thanks


jamesadarich:
Hey @tenowg! Hope all is well with you, just having a look at this pull request and I think I agree with you in that the typings aren't quite right, and definitely should be in the package.json. However I'd like to propose a minor tweak to your commit.

I think it is right to remove the exported module but not the declared one because we need to declare it as something and it also means you can import very neatly either

```
import * as Inversify from "inversify";
```

or specifically what you want

```
import { Kernel, TypeBindings } from "inversify";
```

i.e. 

```
declare module 'inversify'
   export module inversify
```

should become

```
declare module 'inversify'
```

Have tried this on Atom (which I believe VSC is based upon) on some of my solutions and this seems to be working quite neatly. What do you think? :)

I believe this is what @pixelshaded was talking about in #18 so my thoughts are that this PR should close it (with the above change)

Also as a side note don't be worried about deleting instead of commenting stuff out! Things are neater that way and we can always revert mistakes :)

Nice one on the contribution!!!

tenowg:
@jamesrichford you are correct, I had it backwards... using declare module 'inversify' allows for much smoother imports

tenowg:
Seems I was in error... and not sure why... I am going to investigate this more, but doing the last commit makes this fail with visual studio, with "not a module" error on build, but not in IDE. Let me look into this more before merging

tenowg:
At the chance of sounding a fool, I didn't test the last commit before pushing it... I was wrong in that adding back in `declare module 'invserify'` was required, it isn't required, and is actually what is breaking VS build process. I am not sure if this is a Bug with VS, as the import statement:

```
import {Kernel} from 'inversify';
```

doesn't show an error in the IDE, but only when building does it cry. But after further review, including examining other repos, none seem to use `declare module '...'` syntax. So i am starting to be inclined to believe that it isn't required, as my project also builds with no issue now (yes this time ti tested it with a full build and run) using either import method

```
import * as i from 'inversify';
```

and

```
import {Kernel} from 'inversify';
```

So as of now, with my current understanding, I believe leaving the declare statement out might be acceptable.

jamesadarich:
Just tried this on VS2015 and not getting the same issue - is this perhaps to do with moduleResolution? I have mine set to "node" in my compilerOptions.

tenowg:
Doing some research I found that these are the requirements to use the typings in a package.json

be a .d.ts file
be an external module
not have triple-slash references

So, in order for what I want to do, I will require an additional .d.ts file (probably easiest to achieve by setting definitions to true in tsconfig) or a copy of the one used in type_definitions which would be left alone if someone doesn't want to use moduleResolution... I will fix this one last time and see if it is acceptable.

jamesadarich:
Hey @tenowg!

Have you got a repo with an example of how the imports are happening? as I cannot recreate the issue at the moment (with my updated typings) and would like to because I think you are right that there needs to be a "typings" property in the package.json that references the type_definitions/inversify.d.ts file my only concern is maintaining two versions of the typings.

Looking at my projects I'm sure you are right that the typings file is not quite correct and needs to be changed but probably just as simple as removing the encapsulating export module as above, as after that I can import Inversify as you require.

Cheers :)

tenowg:
there is a repo at https://github.com/tenowg/inversifytest that is setup so all you should need to do is pull it and run tsc from the base directory - if you don't have asp.net 5 installed you shouldn't need to load the project.

I have included the node_modules directory, with a modified version of inversify that adds typings, and points it to a modified version of the inversify.d.ts under type_definitions.

I hope you find the issue, maintaining 2 copies is a pain, but you shouldn't have to maintain two if this works... hopefully is might be an issue with my setup

two things: importing works, if you look at test.ts there is no error, when I run tsc I get the error

tenowg:
I would love to maybe talk about this, I will be watching gitter when I get home from work, about 7 hours from now, if needed

remojansen:
We should be able to import InversifyJS as a:

### Global

```
/// <reference path='typings/inversify/inversify.d.ts' />
var Kernel = inversify.Kernel;
var TypeBinding = inversify.TypeBinding;
var TypeBindingScopeEnum = inversify.TypeBindingScopeEnum;
var Inject = inversify.Inject;
```

### ES6 module

```
import * as inversify from "inversify";
var Kernel = inversify.Kernel;
var TypeBinding = inversify.TypeBinding;
var TypeBindingScopeEnum = inversify.TypeBindingScopeEnum;
var Inject = inversify.Inject;
```

```
import { Kernel, TypeBinding, TypeBindingScopeEnum, Inject } from "inversify";
```

### External module

```
import inversify = require("inversify");
var Kernel = inversify.Kernel;
var TypeBinding = inversify.TypeBinding;
var TypeBindingScopeEnum = inversify.TypeBindingScopeEnum;
var Inject = inversify.Inject;
```

We will only merge the pull request once we get all these options to work. 

Based on the documentation [here](https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages):

> Your typings file should...
> - be a .d.ts file
> - be an external module
> - not have triple-slash references
> -  not pollute global scope
> 
> The rationale is that typings should not bring new compatible sources to the set of compiled files; otherwise source files (i.e. .ts files) will be considered by the compiler as part of the user code and will be compiled, and outputs in the package can be overwritten with the resulting .js outputs.
> 
> Additionally, loading typings should not pollute global scope by bringing potentially conflicting entries from different version of the same library. Modules have their own scope, and do not pollute the global namespace, if your typings file is not a module, it will be polluting the user global scope, and will cause conflicts with other packages that depend on your package. Similarly /// <references ... /> can bring global declarations into the global scope and should be avoided.

It looks like we are going to need two files based on [this issue](https://github.com/TypeStrong/ntypescript/issues/15). 

Take a look to the [React type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/react) there are two files one to be used as a global and one to be used for modules. 

So we need two files:
- `type_definitions/inversify.d.ts`
- `type_definitions/inversify-global.d.ts`

And two files for testing those:
- `type_definitions/inversify-test.ts`
- `type_definitions/inversify-global-test.ts`

The `typings` field in the `packaghe.json` should point to the `inversify.d.ts` file not the  `type_definitions/inversify-global.d.ts`.

basarat:
ES6 import and require import will was thanks to `package.json` + `inversify.d.ts` found in this PR. 

## global

```
> /// <reference path='typings/inversify/inversify.d.ts' />
```

If you just change this to `node_modules/inversify/type_definitions/inversify.d.ts` Then the global scenario will work as well :rose:

remojansen:
@basarat thanks a lot for taking a look :)

tenowg:
Not sure if this is really even right :smile: but I have made the d.ts files work in all instances requested, and leaving the need to maintain only one d.ts file. Please review and let me know. thanks

note: the error is likely because there isn't a base tsconfig.json file on the project (at least this is the case in VSC)

remojansen:
Hi guys thanks for working on this and sorry it took so long to merge the changes...

For future references:

Use `inversify-global.d.ts` if you are going to use Inversify as a global:

```
/// <reference path="inversify-global.d.ts" />

var kernel = new inversify.Kernel();

// ...
```

Use `inversify.d.ts if you are going to use InversifyJS as a module:

```
/// <reference path="inversify.d.ts" />

import { TypeBindingScopeEnum, TypeBinding, Kernel, Inject } from "inversify";

var kernel = new Kernel();

// ...
```

PR #39, opened 2016-01-20
Fix #37 and #40: Bloated npm distribution, use Mocha to run tests

### LINKED ISSUE #37: Bloated npm distribution
At the moment, too much is being distributed with the npm distribution.

Included in the distribution is the TypeScript source, the Karma configuration, and the test files.

The package size could be slimmed down considerably by only distributing the compiled JS and the package.json.

Jameskmonger:
I've attached individual commit messages to each of the commits explaining the changes.

I also made a few changes to prevent bloating of the git repo as well as the npm package.

To summarise:
- Create .npmignore file ignoring non-distribution files so that they are not distributed with the package.
- Remove distribution code from the git repo (it should be produced locally from the source)
- Remove typings files from the git repo (all the information needed for the typings is in tsd.json. We should distribute this in the repo rather than all the typings themselves)
- Remove .coveralls.yml from the git repo (this is a security risk, see http://coveralls.io)
- Compile all individual .gitignore files into one single file


Jameskmonger:
This is showing as a failed test because the test coverage is 99.78% - it was at the same percentage before and has not changed due to my changes. Just FYI @remojansen

Jameskmonger:
This addresses #37

Jameskmonger:
I have just taken a look and the current npm distribution is **130kb**. After the initial changes, I have got this down to **46kb**.

Jameskmonger:
I have now got this down to **40kb**.

remojansen:
Hi @Jameskmonger thanks for the pull requests I think the build is failing because coveralls needs a token which was available in the file `.coveralls.yml` I think the token is not needed anymore but we might need to add `service_name: travis-ci` like for example at https://github.com/piwik/device-detector/blob/master/.coveralls.yml can you try that please?

Jameskmonger:
Hey @remojansen, I have had a look and the build is building correctly but is marking as failed for a reason:
- The coverage is not yet at 100%
- This PR has not increased the coverage towards 100%

If a coveralls check does not increase the coverage towards 100%, then coveralls will say that the build has failed. If you take a look at some of your previous builds, https://coveralls.io/builds/4754691 or https://coveralls.io/builds/4754690, you can see that they are also sitting at 99.78%.

I will now add the `service_name: travis-ci` to the `.coveralls.yml` file to see if this makes the build pass.

Jameskmonger:
This is an issue with coveralls @remojansen , please see here: https://github.com/lemurheavy/coveralls-public/issues/545

Jameskmonger:
@remojansen  I have inspected the coverage logs and it seems that the only part not under test is this part:
`var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f`

This is caused where the browserify code is not being tested. I recommend using jasmine-node rather than karma to run the tests, because then we can test the `commonjs` code rather than the `amd` code.

Jameskmonger:
All fixed, @remojansen. By using Mocha we don't have to browserify the code before testing it - we can test the raw commonjs code which means that we're testing the Inversify code by itself rather than after being browserified. It also means that we get accurate coveralls reports rather than them being affected by the browserification process.

And now I can sleep! :stuck_out_tongue_winking_eye:

jamesadarich:
:+1: looks pretty good so far. At a quick skim maybe in .gitignore put comments for the added stuff for consistency and readability? Same for .npmignore as well. But that's super minor so nice rework. Also probably worth hitting bower for the same as people using bower don't deserve to be penalised I guess :stuck_out_tongue_winking_eye: You can do this by adding an ignore property to the bower.json (see here http://themestent.github.io/front-end-workflow/bower.json.html)

remojansen:
:+1: Awesome work!

PR #57, opened 2016-01-28
Converting from tsd to typings

tenowg:
There were no errors in any of the files with imports, this seems a fairly straight forward change... please review and merge if it looks good.


remojansen:
@tenowg thanks a lot for that!

PR #66, opened 2016-02-03
Fix #65: Test and distribute before publishing

### LINKED ISSUE #65: Add prepublish npm hook
Hey @Jameskmonger,

As discussed we should be running automated scripts prepublish to ensure breaking builds cannot be published.

For now just run `gulp test` and we will add more checks as necessary to ensure InversifyJS is as stable as possible.

Jameskmonger:
This automatically runs `gulp test`, followed by `gulp dist` before an `npm publish`, and will not publish if there is any issue with either of those two tasks.

I have tested this locally and it will not publish with failing tests.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -112,7 +112,7 @@ gulp.task("cover", function() {
 });
 
 gulp.task("test", function(cb) {
-  runSequence("istanbul:hook", "mocha", "cover", cb);
+  runSequence("build", "istanbul:hook", "mocha", "cover", cb);
 });
 
 //******************************************************************************
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -111,7 +111,7 @@ gulp.task("cover", function() {
       .pipe(coveralls());
 });
 
-gulp.task("test", function(cb) {
+gulp.task("build-and-test", function(cb) {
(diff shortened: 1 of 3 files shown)


Jameskmonger:
This addresses #65

remojansen:
Thanks for the PR :smile: 

I just noticed a small problem. If you take a look to [the default gulp task](https://github.com/Jameskmonger/InversifyJS/blob/master/gulpfile.js#L166-L167) you will see that it runs the `build` and `test` tasks. The default task is executed by Travis CI so the `build` task [will be executed two times](https://travis-ci.org/inversify/InversifyJS/jobs/106616766#L1018-L1033).

Can you please rename the `test` task as `build-and-test`:

```
gulp.task("build-and-test", function(cb) {
  runSequence("build", "istanbul:hook", "mocha", "cover", cb);
});
```

And replace the `build` and `test` sub-tasks from the `default` task with the new `build-and-test` task:

```
gulp.task("default", function (cb) {
  runSequence(
    "lint",
    "build-and-test",
    "dist",
    cb);
});
```

jamesadarich:
LGTM CR Passed :fireworks:

remojansen:
Looks good to me too :+1:  Thanks a lot for the PR and CR guys!

Issue #138, opened 2016-03-23
Create an es2015 ready build


hourliert:
Hi,
It would be nice if this library could be exported as an es2015 library. (In that way, it could work out of the box with `webpack2` or `rollup`).
## Expected Behavior
- Keep the `dist` folder containing `inversify.js`.
- Create new folders (`es` or `lib` for instance) including transpiled typescript files without concatenating them.
## Possible Solution

We could take inspiration on [redux](https://github.com/reactjs/redux) which transpile ES2015 to ES5 via babel in 2 separates folders `lib` and `es`. 
I have also created a project [builder-react-comp](https://github.com/hourliert/builder-react-comp) which transpile my TS react component into 2 versions. One compatible with webpack2 and rollup and one compatible with webpack1 or browserify.
An use case of this project is available [here](https://github.com/hourliert/react-component-seed).
## Context

The main issue if we don't do that is that when you are building a project that has a dependency on inversify with `webpack1`. You get this compilation warning:

```
WARNING in ./~/inversify/dist/inversify.js
Critical dependencies:
7:482-489 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/inversify/dist/inversify.js 7:482-489
```

Thanks in advance.

remojansen:
Thanks a lot for reporting this issue. I have one question. I like the Redux style:
- dist (bundled + sourcemaps and compressed bundled + sourcemaps)
- lib (es5 not compressed)
- es (es6 not compressed)

What about sourcemaps for lib and es6? Should they be included? I took a look top redux and it doesn't include sourcemaps (even for the compressed file in dist).

Thanks!

hourliert:
Hi,

You're right. I have never noticed that (maybe because I never had to debug something related to redux).
I guess, we still could include source maps like in the dist folder as long as they are not inline.

We could use the typescript compiler directly (without using gulp) to output these 2 folders. Or just create new gulp tasks to be standardized.

What do you think? I could do the PR if you want.

remojansen:
I think is good idea two add them as a gulp task. I was actually thinking about adding an extra compilation targeting es6 a couple of days ago. 

I wasn't thinking about including the output to the npm releases, I was thinking about catching potential duplicated definitions errors caused by the `es6.d.ts` file that is included by the compiler when we target es6.

We are including sourcemaps already in the dist folder using `.map` files. So we just need:
[...]

PR #139, opened 2016-03-23
Add typings to package.json

hourliert:
This allows Typescript to automatically resolve types definition files.

No more need to use typings via `DefinitelyTyped` or to import it via a `/// <ref...`.
## Types of changes
## Checklist:
- [x] My change requires a change to the type definitions.
- [x] I have updated the type definitions accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ npm install inversify@2.0.0-alpha.3 --save
 The InversifyJS type definitions are included in the npm package:
 
 ```ts
-/// <reference path="node_modules/inversify/type_definitions/inversify.d.ts" />
+/// <reference path="node_modules/inversify/type_definitions/inversify/inversify.d.ts" />
 ```
 > **Note**: InversifyJS requires a modern JavaScript engine with support for the Promise, Reflect (with metadata) and Proxy objects. If your environment don't support one of these you will need to import a shim or polypill. Check out the [Environment support and polyfills](https://github.com/inversify/InversifyJS/wiki/Environmemt-support-and-polyfills) page in the wiki to learn more.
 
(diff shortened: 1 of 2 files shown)


remojansen:
I'm sorry but I cannot merge this. There is a problem with the `typings` field in the `package.json` file. We tried to used a few weeks ago but we had to roll it back.

You can read [here](https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages#your-typings-file-should) about the requirements:

> Your typings file should...
> - be a .d.ts file
> - be an external module
> - not have triple-slash references

We meet the first two requirements but we [have a dependency](https://github.com/inversify/InversifyJS/blob/master/type_definitions/inversify/inversify.d.ts#L6-L7) on `bluebird.d.ts` and `harmony-proxy.d.ts` which means that we don't meet the third requirement and therefore we cannot use the `typings` field in the `package.json` file.

Right now we are in alpha so we are just explaining how to to add a manual reference but in the future we are going to add the type definitions to the typings registry. If you know how to do this better your help is more than welcome :smile: 

If you want to update the PR to send just the `README.md` changes I will be happy to merge it.

hourliert:
Yep, I have just noticed that when testing the PR for https://github.com/inversify/InversifyJS/issues/138.

Moreover, I haven't noticed that the type definitions files for Inversify are not auto-generatred by Typescript. Is there a reason against that?

remojansen:
We created them manually because some people was having problems. The problem was that we couldn't find a way to create one single type definition file that would work all the possible ways of consuming it:
- Add `// <reference>`
- No `// <reference>` but typings in package JSON
- es6 imports `import { } from "inversify"`
- typescript external module `import inversify = require('inversify');`
- global `window.inversify`

hourliert:
I see. Seems legit!

PR #173, opened 2016-04-11
Fixes #168

### LINKED ISSUE #168: False positive: Derived class must explicitly declare its constructor.
Hi,

Since https://github.com/inversify/InversifyJS/pull/158, I have a false positive when the base class doesn't use DI and the derieved class does.
## Expected Behavior

We should be able to extend class that doesn't use DI (and is not annotated by `@injectable()`) to create a new class that is `@injectable()` and uses DI.
## Current Behavior

The following snippet doesn't work:

**Base Class:**

``` ts
class Store<T> {
  constructor(
    private _config: T
  ) {}
}
```

**Derived Class:**

``` ts
@injectable()
class ServerStore extends Store<IServer> {
  constructor() {
    super({ name: 'C9000-AWS' });
  }
}
```

When I `@inject()` `ServerStore`, I have this error:
`Error: Derived class must explicitly declare its constructor: ServerStore.`
## Possible Solution

We should reinforce [this condition](https://github.com/remojansen/InversifyJS/blob/master/src/planning/planner.ts#L222).

I have a lot of work right now. But I guess I will be able to fix this Monday or Tuesday?
## Steps to Reproduce (for bugs)

Check the previous snippet.
## Your Environment
- Version used: **beta.1**
# Stack trace

```
Error: Derived class must explicitly declare its constructor: ServerStore.
    at Planner._createSubRequest (/Users/thomashourlier/Projects/react/retax/node_modules/inversify/lib/planning/planner.js:73:23)
    at /Users/thomashourlier/Projects/react/retax/node_modules/inversify/lib/planning/planner.js:27:60
    at Array.forEach (native)
    at Planner.createPlan (/Users/thomashourlier/Projects/react/retax/node_modules/inversify/lib/planning/planner.js:27:22)
    at Kernel._planAndResolve (/Users/thomashourlier/Projects/react/retax/node_modules/inversify/lib/kernel/kernel.js:116:23)
    at Kernel._get (/Users/thomashourlier/Projects/react/retax/node_modules/inversify/lib/kernel/kernel.js:108:25)
    at Kernel.get (/Users/thomashourlier/Projects/react/retax/node_modules/inversify/lib/kernel/kernel.js:60:21)
    at Object.<anonymous> (/source/src/index.ts:20:39)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
```

hourliert:
This fixes #168.

Inversify will raise the `MISSING_EXPLICIT_CONSTRUCTOR` error only if the base class is decorated by `@injectable()` and the user has forgotten to redeclare the constructor in the newly derived class (same condition introduced with https://github.com/inversify/InversifyJS/pull/158).
## Description

Since https://github.com/inversify/InversifyJS/pull/158, Inversify always raises an error when the user extends a class without redeclaring the constructor and injects this derived class 'somewhere' . This could lead to false positive.

For instance, the user could simply extends a class that has one parameter (that doesn't need to be injected).

**Case 1.**

``` ts
class Samurai implements ISamurai {

    public rank: string;

    public constructor(rank: string) {
        this.rank = rank;
    }
}

@injectable()
class SamuraiMaster extends Samurai implements ISamurai {
    constructor() {
        super("Master");
    }
}
```

``` ts
let samurai = kernel.get<ISamurai>("ISamuraiMaster");
expect(samurai.rank).eql("Master");
```

On the other side, the user could extend a class that has one parameter (that should be injected).

**Case 2.**

``` ts
class Samurai implements IWarrior {

    public weapon: IWeapon;

    public constructor(weapon: IWeapon) {
        this.weapon = weapon;
    }
}

@injectable()
class SamuraiMaster extends Samurai implements IWarrior {
    public isMaster: boolean;
    public constructor(@inject(SYMBOLS.IWeapon) weapon: IWeapon) {
        super(weapon);
        this.isMaster = true;
    }
}
```

This PR makes the error `MISSING_EXPLICIT_CONSTRUCTOR` raiser softer. It will be thrown only if the base class is also decorated with `@injectable()`. (In **case 2**, it means `Samurai` should be decorated with `@injectable()`). 
This will allows both case 1 & 2 to be valid.
## Related Issue

https://github.com/inversify/InversifyJS/issues/168
## Motivation and Context

https://github.com/inversify/InversifyJS/issues/168
## How Has This Been Tested?

Unit tests
## Types of changes
## Checklist:

- [ ] My change requires a change to the type definitions.
- [ ] I have updated the type definitions accordingly.

### DIFF
--- a/src/planning/planner.ts
+++ b/src/planning/planner.ts
@@ -91,8 +91,8 @@ class Planner implements IPlanner {
 
             } else {
 
-                // one ore more than one matching bindings found 
-                // when more than 1 matching bindings found target is an array 
+                // one ore more than one matching bindings found
+                // when more than 1 matching bindings found target is an array
                 this._createChildRequest(parentRequest, target, activeBindings);
 
             }
@@ -188,7 +188,7 @@ class Planner implements IPlanner {
             let multiInject: any = targetMetadataMap[METADATA_KEY.MULTI_INJECT_TAG];
             let targetName: any = targetMetadataMap[METADATA_KEY.NAME_TAG];
 
-            // Take type to be injected from user-generated metadata 
+            // Take type to be injected from user-generated metadata
             // if not available use compiler-generated metadata
             targetType = (inject || multiInject) ? (inject || multiInject) : targetType;
 
@@ -207,7 +207,7 @@ class Planner implements IPlanner {
         }
 
         // Throw if a derived class does not implement its constructor explicitly
-        // We do this to prevent errors when a base class (parent) has dependencies 
+        // We do this to prevent errors when a base class (parent) has dependencies
         // and one of the derived classes (children) has no dependencies
         if (targets.length === 0 && this._baseClassHasDepencencies(func)) {
             throw new Error(`${ERROR_MSGS.MISSING_EXPLICIT_CONSTRUCTOR} ${constructorName}.`);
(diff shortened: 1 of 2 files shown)


remojansen:
Thanks a lot :) I like the way you solved the issue :+1:

PR #253, opened 2016-06-18
Added typings instructions

davidstellini:
Updated readme.md


remojansen:
Thanks a lot for the PR but I won't be able to merge. Sorry if the current installation process is not as easy as one would expect but we are working on [the new .d.ts distribution system](https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/). This means that typings will not be required in the future. This is why we started using `inversify-dts`. I hope this doesn't put you off contributing in the future.

PR #284, opened 2016-06-30
[WIP] Improving the installation process

remojansen:
WIP ignore for now...


remojansen:
I will not merge this I've been trying and [even angular 2.0 avoid including reflect-metadata](https://angular.io/docs/ts/latest/quickstart.html#!#add-config-files). You can see that it is imported [only once as a global](https://angular.io/docs/ts/latest/quickstart.html#!#index) I know it is a pain to deal with. I will work on improving the docs as much as I can bit not change the way it is imported.

PR #346, opened 2016-08-23
Doc: Update get_all_named_tagged.md

Offirmo:
Minor doc update.


lholznagel:
I hink the `[]` needs to stay there:
`let weapons = kernel.getAll<Weapon[]>("Weapon");  // returns Weapon[]`

Offirmo:
@lholznagel I tested it for real. Using [] cause a typescript failure claiming a return type of Weapon[][]. Not using [] works as expected.

It also makes sense. The [] is redundant and adds nothing.

lholznagel:
Oh okay, thanks for trying :)

remojansen:
Thanks for this as well :)

PR #380, opened 2016-09-23
clean up the entry point

goenning:
## Description

I was running through the code and thought that it'd be better to remove duplicates from the entry point by making it single line exports.
## How Has This Been Tested?

`npm test` passed


remojansen:
Thanks for the PR 😄

remojansen (commented):
👍

PR #400, opened 2016-11-04
[docs] Map is now required

goenning:
Just saw that v3 landed on master. Might be good to warn devs that Maps are now required and not all browsers (<= IE10) support it.


remojansen:
Awesome thanks a lot for this PR 👍

PR #458, opened 2016-12-24
Fix uppercase

thinkingmedia:
Fix uppercase for `BindingScope`

```
    type BindingScope = "Singleton" | "Transient";
```


remojansen:
Thanks for the PR 👍

Issue #462, opened 2016-12-28
Auto override bindings feature.


ivan-shaban:
I dont found any information about how to allow auto override exist binding, but i think it will be usefull.

## Expected Behavior

I load several modules and if some of them has the same binding IDs, so only last binding became active.

## Current Behavior

I load several modules and if some of them has the same binding IDs, i got a runtime error.

## Possible Solution

E.g. something like:
```
public bind<T>(serviceIdentifier: interfaces.ServiceIdentifier<T>, overrideIfExists:Boolean = false): interfaces.BindingToSyntax<T> {
```
In [SwiftSuspenders library](https://github.com/tschneidereit/SwiftSuspenders) was inverted behaviour, you may override everything at any time and, to prevent it, you should manually `lock` binding with explicit call, e.g.
```
injector.map("serviceIdentifier").toValue("my-value").seal();
```

[In this blob](https://github.com/tschneidereit/SwiftSuspenders/blob/master/src/org/swiftsuspenders/mapping/InjectionMapping.as) in function `seal` described how it works

## Steps to Reproduce (for bugs)

1. Make 2 modules with the same serviceIdentifiers
2. Load them into one container

## Context

I develop feature to extend application with 3rd party plugins, every plugin represented as self-independ module and all plugins loads into one container. Because plugins will be developed by different people it is possible that they may override the same `serviceIdentifier`s in that case i wanna to show warning message that plugins are not compatible, but don't want to block appllication work, it will be enough if only last binding override stay active. 

## Your Environment

    "inversify": "^3.0.0-rc.2",
    "reflect-metadata": "^0.1.9",

# Stack trace

remojansen:
Hi @LordXaoca thanks for using inversify and for your suggestion 👍 

Instead of sealed bindings we allow you to add and remove bindings. You can remove a binding and then create a new one using [`container.unbind`](https://github.com/inversify/InversifyJS/blob/master/src/container/container.ts#L127).

At the moment we only allow you to remove all bindings for a given service identifier but there is already an issue to improve this #447

remojansen:
Re-opening issue as the requirement is to allow to invoke unbind from container modules.

remojansen:
I have implemented support for unbind within container modules https://github.com/inversify/InversifyJS/pull/464 It will be released over the next few days

PR #464, opened 2016-12-29
Support for unbind within container modules

remojansen:
## Description
Support for unbind within container modules

## Related Issue
#462

## Motivation and Context
#462

## How Has This Been Tested?
Added and unit test and tested using CI

## Types of changes

## Checklist:


ivan-shaban:
I checked patch and i think it is better to implement in different way. Yes, for now i will able to unbind something in module, but how do i know that i should do that? I don't see any ability to check is there binding exists( So i propose to pass in module optional "override strategy" function, it will solve override situations, for now i see possible scenarios:
- throw error is binding exists (default)
- override old binding with new binding (useful when you wanna update config)
- skip new binding if old exists (to prevent update, but do not throw error)

There are probably more strategies or user event may write own strategy if he need it. What do you think about it?

remojansen:
Hi @LordXaoca about:

> I don't see any ability to check is there binding exists

I will be solved that problem by providing access to `isBound` to the modules. I will do this later today. Check out [the wiki](https://github.com/inversify/InversifyJS/blob/master/wiki/container_api.md#containerisbound) to learn more.

I don't want to add sealed bindings for now because there is already a way to achieve the same with: `bind`, `unbind` and `isBound`. I think it gets messy when a library offers many ways to do something. At the same time I listen to the user's feedback and I really appreciate your feedback. If more users mention the need for this I will implement it.

ivan-shaban:
Hi, sorry for late answer, in comment above i'm not talking about sealed bindings, i'm talking about bindings management. I check tests and, i think, i understand how library works, as i understand in container configuration i can make several binginds with the same `serviceIdentifier` (e.g. to use `Multi-injection`), but formally i get error only when i try to get bad-mapped-instance. IMHO it is bad behaviour, because it is hard to understand when i really make mistake, probably i manually type wrong binding, or mb i have few modules where the same `serviceIdentifier` was used, dunno. But if we will check bindings when we configure container it will be more explicit, what do you think about it?

PR #465, opened 2016-12-30
resolveInstance should create writable property

mertdeg:
Without this change, sometimes, I get following error:

TypeError: Cannot assign to read only property 'x' of #<Object>

More reading about this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Read-only


remojansen:
Hi @mertdeg thanks for the PR sorry but we already fixed this issue but it has not been released until tonight. The version 3.0.0-rc.4 has been released just now should solve this problem for you.

Issue #468, opened 2017-01-02
Property Injection is not working.


ajaysinghj8:
In case of property Injection inside a class, I am getting an error as 
    instance[propertyName] = injection;
                               ^
TypeError: Cannot assign to read only property 'logger' of object '#<Router>'.

## Your Environment
 "inversify": "^3.0.0-rc.2",
  "node": "7.2.0",
  "typescript":"2.2.0"

# Stack trace
/Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:17
        instance[propertyName] = injection;
                               ^
TypeError: Cannot assign to read only property 'logger' of object '#<Router>'
    at /Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:17:32
    at Array.forEach (native)
    at _injectProperties (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:10:32)
    at Object.resolveInstance (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:34:18)
    at _resolveRequest (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/resolver.js:48:38)
    at /Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:8:16
    at Array.map (native)
    at _injectProperties (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:7:57)
    at Object.resolveInstance (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:34:18)
    at _resolveRequest (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/resolver.js:48:38)
    at Object.resolve (/Users/abc/xyz-project/node_modules/inversify/lib/resolution/resolver.js:65:12)
    at Container._planAndResolve (/Users/abc/xyz-project/node_modules/inversify/lib/container/container.js:193:37)
    at Container._get (/Users/abc/xyz-project/node_modules/inversify/lib/container/container.js:184:44)
    at Container.get (/Users/abc/xyz-project/node_modules/inversify/lib/container/container.js:149:21)
    at Object.<anonymous> (/Users/abc/xyz-project/dist/modules/server/server.spec.js:14:26)
    at Module._compile (module.js:571:32)

# Possible Solution 
/Users/abc/xyz-project/node_modules/inversify/lib/resolution/instantiation.js:17
        instance[propertyName] = injection; 
[...]

PR #469, opened 2017-01-02
bugfix: #468

ajaysinghj8:
Property Injection is not working. #468


remojansen:
This should be already fixed in `3.0.0-rc.4` can you please confirm? If you are using `inversify@2.X.X` you will need `reflect-metadata@0.1.8` instead of `reflect-metadata@0.1.9`.

Issue #587, opened 2017-06-27
add @postConstruct to the lifecycle


dcavanagh:
I have added @postConstruct method decorator that will run after an object is instantiated like in [Spring](https://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#beans-postconstruct-and-predestroy-annotations)

## Expected Behavior
if you decorate a method with @postConstruct that method will be called after the object is constructed

## Current Behavior
This behavior does not currently exist

## Possible Solution
I am working on a change and have it working but it needs some review and maybe some refactoring to make it fit in inversify better before i submit a pull request. [fork](https://github.com/AltekkeE/InversifyJS)

## Steps to Reproduce (for bugs)
N/A

## Context
Keep object setup out of the constructor

## Your Environment

* Version used: 4.1.1
* Environment name and version (e.g. Chrome 39, node.js 5.4): Node v7.4.0
* Operating System and version (desktop or mobile): OSX
* Link to your project:  [fork](https://github.com/AltekkeE/InversifyJS)

# Stack trace
N/A

remojansen:
Hi @AltekkeE I like this feature :+1: I have some questions if you don't mind... 

How do you think this should play with the `onActivation` feature? should  `@PostConstruct`  be invoked before or after the `onActivation` handler? 

I'm thinking that maybe after because we could have the `onActivation` for all implementations of a Type and `@PostConstruct` for one of the implementations. Do you agree?

dcavanagh:
I agree! `@postConstruct` is meant to allow for implementation specific setup. I'm not sure about the order since the `onActivation` feature can change the behavior of the class if understand it correctly.

remojansen:
True, thanks for bringing that to my memory. I think `@postConstruct` should go first then. So the order should be:

1. `const x = new X()` // new instance
2. `x.postConstruct()` // instance initialized
3. `onActivationHandler(x)` // 

I also think that we should wrap `postConstruct` with a try catch and if an error takes place throw a custom error. Something like:

> `@postConstruct` error in class X: The details of the original exception goes here.

Do you have any other concerns?

dcavanagh:
Sounds good.. I will add that

My only other concern is that i did not add any code to the annotation phase or the plan phase. 
The resolution phase just checks to see if the `POST_CONSTRUCT` metadata exists on the prototype and then executes the method that was decorated

gagle:
preDestroy could be useful too

[...]

PR #588, opened 2017-06-27
added @postConstruct method decorator

dcavanagh:
decorate a method with @PostConstruct and that method will be called after the object is constructed
## Description

added @PostConstruct method decorator that will run after an object is instantiated like in Spring
added string templating method

## Related Issue
https://github.com/inversify/InversifyJS/issues/587

## Motivation and Context

Keep object setup out of the constructor

## How Has This Been Tested?
Several unit tests were added. I also tested this on my own project

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.  Not sure where this should go
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


remojansen:
Hi @AltekkeE thanks a lot for this PR. Awesome work :+1:  I don't see any problems at the implementation details level. At the architecture level, there is a problem that you have already mentioned @ https://github.com/inversify/InversifyJS/issues/587#issuecomment-311414698:

> My only other concern is that i did not add any code to the annotation phase or the plan phase.

The way I would preffer it to work is the following:

1. Move the reading of the metadata to [metadata_reader.ts](https://github.com/inversify/InversifyJS/blob/93f83c2bcfbb06e2da7fa564c09d9f4ce77a8d38/src/planning/metadata_reader.ts). 

2. The [reflection_utils.ts](https://github.com/inversify/InversifyJS/blob/d66c29c96a2b118714589f50de6296b6c52ba78a/src/planning/reflection_utils.ts) will consume the metadata reader.

3. The plan phase [asks the reflection ustils](https://github.com/inversify/InversifyJS/blob/d60a0539b197ac9c792033db8c735395062f2ff2/src/planning/planner.ts#L169-L173) for the dependencies of an object. Then it [creates a request for each dependency](https://github.com/inversify/InversifyJS/blob/d60a0539b197ac9c792033db8c735395062f2ff2/src/planning/planner.ts#L141-L154). We need to put the post construct handler somewhere I'm thinking that the right place is probably or the`Request` class not the `Binding` class.

3. The resolve phase should then invoke the handler [like here](https://github.com/inversify/InversifyJS/blob/d66c29c96a2b118714589f50de6296b6c52ba78a/src/resolution/resolver.ts#L62-L65). It will look something like:

```ts
        // use post construct handler if available
        if (typeof request.postconstructMethodName !== null) {
            result[request.postconstructMethodName]();
        }

        // use activation handler if available
        if (typeof binding.onActivation === "function") {
            result = binding.onActivation(request.parentContext, result);
        }
```

This is much more code and work but it is needed by [a feature](https://github.com/inversify/InversifyJS/pull/544) that allows to create your own metadata reader. This was needed to allow people to create their own  custom anotation system.

I'm happy to merge this PR now and refactor later? Please let me know if you want to do the refactor.

dcavanagh:
@remojansen Merging it now would be great. I will take a stab at the refactor. 

I also need to address the documentation. I was thinking of adding a "Life cycle hooks" file to the wiki. Would that be appropriate?

I've been having a problem debugging the unit tests in vsCode. Do you use a specific setup for this?

remojansen:
@AltekkeE I have merged it :tada: will try to do a release later tonight.

Adding a wiki page about life cycle hooks sounds like a good idea :+1: 

Debugging unit test is quite a manual process now. I have to do something about it :cry: I can't remember the steps but you basically need to compile the test and then run mocha with debug brk and then plug VSCode...

dcavanagh:
I figured out how to debug the mocha tests in vscode. 
`{
    // Use IntelliSense to learn about possible Node.js debug attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387


    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "launch",
            "name": "Mocha Tests",
            "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
            "args": [
                "--require",
                "reflect-metadata",
                "-u",
                "tdd",
                "--timeout",
                "999999",
                "--colors",
                "${workspaceRoot}/test/**/*.js"
            ],
            "internalConsoleOptions": "openOnSessionStart"
        },
        {
            "type": "node",
            "request": "launch",
            "name": "Launch Program",
            "program": "${workspaceRoot}/lib/inversify.js",
            "outFiles": [
                "${workspaceRoot}/out/**/*.js"
            ]
        }
    ]
}

`

remojansen:
Do you mind sending that as a PR? :+1:

Issue #549, opened 2017-05-06
Maximum call stack size exceeded


akosyakov:
Sometimes i am getting such errors and it takes a while to debug through inversify code and check bindings to find out reasons behind them. Is there a better to investigate them?

remojansen:
Can you please share details about your environment (version of node, ts, inversify, browser, etc.) also can you share a code example that reproduces the issue?

Call stack issues are mostly related with circular dependencies but Inversify should catch those and explain where is the circular dependency located.

marshalYuan:
Maybe in activation handler!
```
container.bind<Ninja>("Ninja").to(Ninja);

container.bind<Katana>("Katana").to(Katana).onActivation((context, katana) => {
    let handler = {
        apply: function(target, thisArgument, argumentsList) {
            console.log(`Starting: ${new Date().getTime()}`);
            let result = target.apply(thisArgument, argumentsList);
            console.log(`Finished: ${new Date().getTime()}`);
            return result;
        }
    };
    katana.use = new Proxy(katana.use, handler);
    return katana;
});
```
This demo will throw the Error when run many times(exceed nodejs max callstack size)

remojansen:
@akosyakov can you please confirm that you are using `onActivation` handler?

remojansen:
@marshalYuan I just tried with the following unit test and it worked fine:

```ts
it("Should not throw maximum call stack size exceeded", () => {

    interface Ninja {
        fight(): string;
    }

    interface Katana {
        hit(): string;
    }

    @injectable()
    class Katana implements Katana {
        public hit() {
            return "cut!";
        }
    }

    @injectable()
    class Ninja implements Ninja {

        private _katana: Katana;

        public constructor(
            @inject("Katana") katana: Katana
        ) {
            this._katana = katana;
        }

        public fight() {return this._katana.hit(); }

    }

    let container = new Container();
    container.bind<Ninja>("Ninja").to(Ninja);
    container.bind<Katana>("Katana").to(Katana).onActivation((context, katana) => {
        let handler = {
            apply: function(target: any, thisArgument: any, argumentsList: any) {
                console.log(`Starting: ${new Date().getTime()}`);
                let result = target.apply(thisArgument, argumentsList);
                console.log(`Finished: ${new Date().getTime()}`);
                return result;
            }
        };
        katana.hit = new Proxy(katana.hit, handler);
        return katana;
    });

    let ninja = container.get<Ninja>("Ninja");
[...]

PR #604, opened 2017-07-10
Reproduce GH-549

akosyakov:


remojansen:
Hi @akosyakov thanks for helping to find this one out 👍  This is now implemented by #666 I was unable to display the dependency graph in the error messages but it shows something like:

> It looks like there is a circular dependency in one of the 'toDynamicValue' bindings. Please investigate bindings withservice identifier 'B'

PR #643, opened 2017-10-18
Delete obsolete documentation on module bundlers

nLight:
The issue with the static require('crypto') has been resolved. No need to keep the doc.

https://github.com/rbuckton/reflect-metadata/commit/094d953deb0fa38188ea588978d9f2bf93e11bd1


remojansen:
Thanks for this PR :+1:

Issue #500, opened 2017-03-03
New Feature: Automatic binding


ssljivic:
Would it be possible to have container automatically built with all the classes marked with @injectable() instead of manually calling the following code for each class?

`container.bind<MyClass>(MyClass).toSelf();`

Dirrk:
See here:

https://github.com/inversify/inversify-binding-decorators/

remojansen:
Hi, thanks for sharing your feedback with us :+1:  

I'm programming languages like Java or C# we are able to use reflection to explore the types in an assembly or DLL. We can then automatically generate the bindings. Unfortunately, in JavaScript we don't have a reflection mechanism as powerful and therefore all we can do is to create some utilities and helpers like the `inversify-binding-decorators`. 

The highest level of auto-registration that we can provide is [the auto-provice utility](https://github.com/inversify/inversify-binding-decorators/#the-auto-provice-utility).

I'm closing this issue. Feel free to add a comment if you have any additional questions.

ps @Dirrk thanks for helping here

PR #651, opened 2017-10-29
Automatic binding for @injectable() decorated classes

### LINKED ISSUE #649: Bringing toSelf() to the next level
I am aware of #500, #505 and inversify-binding-decorators. Nonetheless, I would love to have a container option **fallbackToSelf** at hand.

Motivation:
- not having to type implicit bindings
- same convenience as in angular: only @Injectable() and @Inject() are necessary (bindings are not mandatory)

What it does / How it behaves:
- Whenever the container would otherwise throw "No matching bindings found..." for an @injectable() class, it will use itself as the binding and therefore not throw at that point
- This "fallback" behavior runs in the "planning" phase, meaning after any manual binding definitions
  - @injectable() and **container.bind** will not be affected at all by this container option
  - **container.get** method will need changes (specifically the planning logic)
- @provide decorator of inversify-binding-decorators does not do the job, because the container must be known at "decorating" phase (which I consider an antipattern btw)
- If possible, **fallbackToSelf** will not add a binding at any point, but rather fall back to the "toSelf()" binding if no binding is defined for an @Injectable() class

Some examples:
```ts
@injectable() class A {}
@injectable() class B { constructor(a: A) {} }
@injectable() class AFake {}
class C {}

const container1 = new Container({fallbackToSelf: true});
container1.get(A); // does not throw, returns transient A
container1.get(B); // does not throw, returns transient B

const container2 = new Container({defaultScope: 'Singleton', fallbackToSelf: true});
container2.get(A); // does not throw, returns singleton A
container2.get(B); // does not throw, returns singleton B

const container3 = new Container({fallbackToSelf: true});
container3.bind(A).toSelf().inSingletonScope();
container3.get(A); // returns singleton A because of explicit binding
container3.get(B); // does not throw, returns transient B with singleton A

const container4 = new Container({fallbackToSelf: true});
container4.bind(A).to(AFake);
container4.get(A); // returns transient AFake because of explicit binding
container4.get(B); // does not throw, returns transient B with transient AFake

const container5 = new Container({fallbackToSelf: true});
container5.get(C): // still throws -> No matching bindings found
```
I tried to do this somehow with a middleware, but I think it is not possible.

osi-oswald:
## Description

Fixes #649

## Related Issue

#500

## Motivation and Context

See #649

## How Has This Been Tested?

See #649 and new tests

## Types of changes

## Checklist:


osi-oswald:
Maybe the option should be called **autoBindInjectable** instead, not sure...

remojansen:
Thanks for this PR 👍

Issue #691, opened 2017-11-24
Use Symbol.for(key) instead of Symbol(key) everywhere


remojansen:
## Expected Behavior
We should use in all our code / examples / docs (and recommend using) `Symbol.for(key)` to create new symbols instead of `Symbol(key)`  because it is unsafe!

## Current Behavior
We currently use in all our code / examples / docs (and recommend using) `Symbol(key)` to create new symbols. This is unsafe!

## Steps to Reproduce (for bugs)
The following code snippet demonstrates why it is unsafe:
```
ts-node temp.ts 
```

####

```ts
function test1() {
    const m = new Map<symbol, number>();

    for (let i = 0; i < 50; i++) {
        const key = Symbol.for(`KEY_${i}`);
        const value = i;
        m.set(key, value);
    }

    for (let i = 0; i < 50; i++) {
        const key = Symbol.for(`KEY_${i}`);
        const value = m.get(key);
        if (value === undefined) {
            throw new Error(`Cannot get ${key.toString()}`);
        }
    }

    console.log("TEST 1 OK :)");
}

function test2() {
    const m = new Map<symbol, number>();

    for (let i = 0; i < 50; i++) {
        const key = Symbol(`KEY_${i}`);
        const value = i;
        m.set(key, value);
    }

    for (let i = 0; i < 50; i++) {
        const key = Symbol(`KEY_${i}`);
        const value = m.get(key);
        if (value === undefined) {
            throw new Error(`Cannot get ${key.toString()}`);
        }
    }

    console.log("TEST 1 OK :)");
}

test1();
test2();
```

## Context
Using `Symbol(key)` is unsafe because the symbols are not unique!

From [MDN about `Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol):

> To create a new primitive symbol, you write Symbol() with an optional string as its description:
> 
> ```js
> var sym1 = Symbol();
> var sym2 = Symbol('foo');
> var sym3 = Symbol('foo');
> ```
>
> The above code creates three new symbols. Note that Symbol("foo") does not coerce the string "foo" into a symbol. It **creates a new symbol each time**:
> 
> ```js
> Symbol('foo') === Symbol('foo'); // f
> ```
And [MDN about `Symbol.for(key) `](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for):

[...]

PR #708, opened 2017-12-09
Replaced all of Symbol( with Symbol.for(

JoshuaKGoldberg:
## Description

Replaces in-code usages of `Symbol(string)` with `Symbol.for(string)`

## Related Issue

#691 

## Motivation and Context

See issue. `Symbol("foo") !== Symbol("foo")` while `Symbol.for("foo") === Symbol.for("foo")`. We want the latter to help remove hard-to-detect seemingly identical but actually different identifiers.

## How Has This Been Tested?

Still covered by unit tests.

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes. _(existing tests updated)_


remojansen:
Thanks a lot for the PR :+1:

idandrd:
The documentation states that the use of Symbols is meant to prevent collisions. Using `Symbol.for` re-introduces this exact problem, rendering the use of Symbols useless, since it now has the exact same effect as using string literals, just with more code.
Am I missing something here?

haoadoreorange:
I agree with @idandrd,

Come across this part today and realize that the use of Symbols has no effect at all.
Using `Symbol.for` result in naming collision if other libs also use the same name. This can lead to hard-to-find bug and render the whole "Support for Symbos" useless.

Please review it in priority.

Issue #772, opened 2018-01-25
Circular dependency on symbols fails too late


krzkaczor:
If there is a circular dependency on symbols and `@inject` is called with `undefined`, error is thrown too late — during dependency resolution from container (`container.get`).

## Expected Behavior
Inversify should early realize that `undefined` is injected and throw an error. There should be a possibility to provide a function that would lazily return symbol which solves the problem of circular dependency.

## Current Behavior
The error is thrown but too late. Related: https://github.com/inversify/InversifyJS/issues/685

## Possible Solution
https://github.com/inversify/InversifyJS/blob/master/src/annotation/inject.ts#L7 here on before returning function should be a guard check for undefined. I don't believe that anyone would inject `undefined` on purpose. The final solution would be to allow using function lazily returning symbol.

## Steps to Reproduce (for bugs)

[repro](https://github.com/krzkaczor/inversify-symbol-circular-dependency-repro)

## Context
My real problem is that the error is thrown too late. This bug slipped through all my tests (both unit and integration) and was triggered only when the user would click the specific button in my application (we use inversify with react) which would kick action with few dependencies resolved by inversify.

## Related: https://github.com/inversify/InversifyJS/issues/685

@remojansen I am happy to work on PR, I just want to discuss possible solutions first.

remojansen:
Closing this issue because it has been fixed by #774 and it has already been released.

PR #774, opened 2018-01-26
Support using functions returning service identifiers with @inject to avoid circular deps

krzkaczor:
## Description
First commit introduces guard check to prohibit injecting `undefined` values as service identifiers. This could happen when circular dependency between symbols is detected.

Second commit adds a possibility to inject parameterless functions returning service identifiers. Implementation is not as nice as I imagined because `ServiceIdentifier` can itself be a `newable` or `abstract` — i don't really understand this usecase but this is in typings. @remojansen could you point me to some docs? 

Documenting this change is missing. I wanted to discuss this solution first. 

## Related Issue
https://github.com/inversify/InversifyJS/issues/772

## How Has This Been Tested?
I added new test cases. 

Previously it was possible to pass `newable` or `abstract`. Now if parameteless function is detected we invoke it immediately. This is a breaking change. 

## Types of changes

## Checklist:

### DIFF
--- a/src/annotation/inject.ts
+++ b/src/annotation/inject.ts
@@ -4,20 +4,25 @@ import { interfaces } from "../interfaces/interfaces";
 import { Metadata } from "../planning/metadata";
 import { tagParameter, tagProperty } from "./decorator_utils";
 
-export type ServiceIdentifierOrFunc = interfaces.ServiceIdentifier<any> | (() => interfaces.ServiceIdentifier<any>);
+export type ServiceIdentifierOrFunc = interfaces.ServiceIdentifier<any> | LazyServiceIdentifer;
 
-function isSimpleFunction<T>(value: any): value is () => T {
-  return typeof value === "function" && value.length === 0;
-}
-
-function inject(serviceIdentifierOrFunc: ServiceIdentifierOrFunc) {
-  if (serviceIdentifierOrFunc === undefined) {
-    throw new Error(UNDEFINED_INJECT_ANNOTATION);
+export class LazyServiceIdentifer<T = any> {
+  private _cb: () => interfaces.ServiceIdentifier<T>;
+  public constructor(cb: () => interfaces.ServiceIdentifier<T>) {
+      this._cb = cb;
   }
 
+  public unwrap() {
+    return this._cb();
+  }
+}
+
+function inject(serviceIdentifier: ServiceIdentifierOrFunc) {
   return function(target: any, targetKey: string, index?: number): void {
-    const serviceIdentifier = isSimpleFunction<interfaces.ServiceIdentifier<any>>(serviceIdentifierOrFunc) ?
-      serviceIdentifierOrFunc() : serviceIdentifierOrFunc;
+    if (serviceIdentifier === undefined) {
+      throw new Error(UNDEFINED_INJECT_ANNOTATION(target.name));
+    }
+
     const metadata = new Metadata(METADATA_KEY.INJECT_TAG, serviceIdentifier);
 
     if (typeof index === "number") {
--- a/src/annotation/inject.ts
+++ b/src/annotation/inject.ts
@@ -1,9 +1,14 @@
+import { UNDEFINED_INJECT_ANNOTATION } from "../constants/error_msgs";
 import * as METADATA_KEY from "../constants/metadata_keys";
 import { interfaces } from "../interfaces/interfaces";
 import { Metadata } from "../planning/metadata";
 import { tagParameter, tagProperty } from "./decorator_utils";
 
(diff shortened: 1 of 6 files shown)


remojansen:
Thanks for the PR.

## The problem

I'm not sure about the helper isSimpleFunction. If someone uses:

```ts
@inject(SomeClass)
```

It could be a problem:

```ts
class SomeClass { }
function someFunction() { }

function isSimpleFunction<T>(value: any): value is () => T {
    return typeof value === "function" && value.length === 0;
}

console.log(
    isSimpleFunction(someFunction), // true
    isSimpleFunction(SomeClass), // true !!!! <---- FALSE POSITIVE
    isSimpleFunction("SomeClass"), // false
    isSimpleFunction(Symbol.for("SomeClass")), // false
);
```

## The solution

I think we are going to need a class so the users can use:

```ts
@injec(new LazyServiceIdentier(() => SomeClass))
```

The class could be something very simple:

```ts
class LazyServiceIdentier<T> {
    private _cb: () => interfaces.ServiceIdentifier<T>;
    public constructor(cb: () => interfaces.ServiceIdentifier<T>) {
        this._cb = cb;
    }
    public get() {
        return this._cb();
    }
}
```

So the check function can become:

```ts
function isLazyServiceIdentier<T>(value: any): value is () => T {
    return value instanceof LazyIServiceIdentier;
}
```

That should work:

```ts
class SomeClass {}
function SomeFunction() {}

console.log(
    isLazyServiceIdentier(SomeClass), // false
    isLazyServiceIdentier("SomeClass"), // false
    isLazyServiceIdentier(Symbol.for("SomeClass")), // false
    isLazyServiceIdentier(someFunction), // false
    isLazyServiceIdentier(new LazyServiceIdentier(() => SomeClass)), // true
    isLazyServiceIdentier(new LazyServiceIdentier(() => "SomeClass")), // true
    isLazyServiceIdentier(new LazyServiceIdentier(() => Symbol.for("SomeClass"))) // true
);
```

krzkaczor:
@remojansen thanks for a quick response! Yes, I was aware of this problem that it could break existing code :/ Your solution sounds good! I will work on it today.

remojansen:
Great, thanks a lot.

Also if you don't mind, can you please also change:

```ts
function inject(serviceIdentifierOrFunc: ServiceIdentifierOrFunc) {
  if (serviceIdentifierOrFunc === undefined) {
    throw new Error(UNDEFINED_INJECT_ANNOTATION); // <-- Too early to know the target
  }

  return function(target: any, targetKey: string, index?: number): void {
      // ...
  };
}
```

For:

```ts
function inject(serviceIdentifierOrFunc: ServiceIdentifierOrFunc) {
  return function(target: any, targetKey: string, index?: number): void {
      if (serviceIdentifierOrFunc === undefined) {
          throw new Error(UNDEFINED_INJECT_ANNOTATION(target.name)); // <--- Target ready
      }
      // ...
  };
}
```

You will be able to create a more concrete error msg. Something like:

```ts
export const UNDEFINED_INJECT_ANNOTATION = (name: string) => {
    return `@inject called with undefined this could mean that the class ${name} has `
           "a circular dependency problem. You can use a LazyServiceIdentier to  " + 
           "overcome this limitation.";
};
```

krzkaczor:
@remojansen yes! you're right. Initially, I thought that returned function is being called much later. That's why I also "unwrapped" function there to get symbol itself. 

Once I realized that it's called right after I noticed that it doesn't solve the problem of circular dependencies (i also tested it locally with my repro example). So what I did now is that I unwrap lazy wrapper during getting dependencies on subrequest method. I will update PR in a sec.

krzkaczor:
@remojansen done! Let me know what do you think. Now, locally my repro works (After adding `LazyServiceIdentifer`)

remojansen:
@krzkaczor looking great 💯 thanks a lot for the PR, awesome contribution 👍  I will do a release now.

krzkaczor:
@remojansen awesome! Thanks! I am gonna update our app ASAP to use newer version. 

Btw. I would suggest considering making breaking change on next major release to always force using `() => symbol` with `@inject`. This would make `LazyServiceIdentifer` redundant and would avoid circular dependencies by design. This is exactly what `typeorm` does.

remojansen:
Thanks for the suggestion. I have created an entry about this in the wiki https://github.com/inversify/InversifyJS/blob/master/wiki/circular_dependencies.md

csr632:
LazyServiceIdentier dosent't works with property injection: https://github.com/inversify/InversifyJS/issues/944

I fix it in https://github.com/inversify/InversifyJS/pull/1177

PR #777, opened 2018-01-27
Maintain singleton activation state in a snapshot (Fix #715)

### LINKED ISSUE #715: Restore and inSingletonScope behavior
## Expected Behavior
Injecting a singleton value after container.restore() reinitializes it.

## Current Behavior
Singleton values should be cached and not initialized twice.

## Steps to Reproduce (for bugs)
var inversify = require("inversify");
require("reflect-metadata");

const container = new inversify.Container();

container.bind('bla').toDynamicValue(() => {
    console.log('------initializing------');
    return 'bla';
}).inSingletonScope();

container.get('bla');
container.get('bla');

console.log('---------snapshot and restore---');

container.snapshot();
container.restore();

container.get('bla');
container.get('bla');

Gives the following output:

------initializing------
---------snapshot and restore---
------initializing------

Whereas I would expect to get the following result:

------initializing------
---------snapshot and restore--

## Context
We have some singletons that require additional initialization outside the toDynamicValue function (to avoid circular dependencies). After the restore method is called and a singleton is injected again the additional initialization should be run again, which is undesirable. I looked at the inversify code and it seems that singleton values should be preserved across restores, but they are not.

### LINKED ISSUE #721: Refreshing a singleton
## Expected Behavior
As a programmer I want to be able to reinitialize a singleton on demand so that I can test singletons. For example,

container.bind<Logger>('Logger').toDynamicValue(// some initialization code here that depends on the context).inSingletonScope();
container.get('Logger'); // the logger instance is created and cached
// I test the logger

now I change some configuration that is bound inside the container, I want to reinitialize the logger. The only way AFAIK atm is to rebind the logger to the same dynamicValue function and then do container.get(). I would propose something like:
container.getNew('Logger') 
or
container.get('Logger', {forceNew: true})

that would force reinitialization of a singleton

What makes this proposed functionality even more appealing to me is that I'm currently forced to do initialization of the singleton separately due to a circular dependency, so that every time I want to get a new singleton instance I have to do quite a lot of steps.

zontafil:
Fix #715 

Check the activation flag of each binding and preserve it in case of a singleton

## Description

Not sure if this is correct. This is my first bug fix in inversify.
Since a singleton can be already initialized before a snapshot, it seems reasonable to me to preserve its activation flag.

Not sure if the flag should be preserved for other types of bindings

## How Has This Been Tested?

I have add a unit test for the bug. I instantiate a singleton class before and after a snapshot/restore. The class should be instantiated only once in total.

## Types of changes

## Checklist:

- [xI have added tests to cover my changes.


zontafil:
N.B: the commit reference to a wrong issue (#721). Not sure how to fix the reference

remojansen:
@m3l7 thanks for the PR, please don't close it. I don't have enough time to help this week but I will send a PR to your work next week. We can then merge both changes. I would like to tell you how to fix the references issue but I'm not sure right now I will have to experiment a little bit...

zontafil:
@remojansen ok, right now I'm testing the fix in my project. So far no issues, the singletons are restored correctly

zveljkovic:
Will this be included soon into master?

omairvaiyani:
+1 I have resorted to `inRequestScope` instead and added some workaround checks to ensure my class (which should otherwise be a singleton) does not repeat initialisation steps when re-instantiated.  Would very much appreciate this being merged! I also came across the bug during unit testing with snapshots.

drew-r:
Is there anything I can help with to get this merged?

baptLesta:
>Would very much appreciate this being merged! I also came across the bug during unit testing with snapshots.

Same here.
Let me know if you need any help to get this merged 👍🏻

dcavanagh (approved):

PR #783, opened 2018-02-02
Typo fix in container_api.md: "even/when" if

JoshuaKGoldberg:


remojansen:
Thanks for the PR 🌈 👍

Issue #881, opened 2018-05-24
Project abuses Math.random for GUID  generation


artemdanylenko:
There are many places where GUIDs are used, but their creation - calls to Math.random is computationally expensive on very low-end devices. My understanding from use cases how GUIDs are used is that the only requirement for them is that they need to be unique. This can be easily achieved with an auto-incremented number.

Am I missing any reason why 36 character GUIDs are used?

Would you accept a PR to replace GUID with id which will be just auto-incremented?

artemdanylenko:
Merged with PR above.

PR #882, opened 2018-05-24
Change GUID to incremented counter for better performance

artemdanylenko:
## Description
In our project, we have a big number of dependencies and we need to support quite old devices (for example chromecast v1). We figured out that this change gives significant improvement in performance on these devices, as Math.random is quite slow there.

## Related Issue
#881 

## Motivation and Context
Math.random is slow

## How Has This Been Tested?
All tests pass, tested in our project

## Types of changes

## Checklist:


tiagoschenkel:
I understand the necessity of this improvement, but I would like to point out that this merge request caused a break in the **Public API**, since the method **guid** was renamed to **id**.

Following basic [semver guidelines](https://semver.org), this change generates a new **major** release:

> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards-compatible manner, and
> 3. PATCH version when you make backwards-compatible bug fixes.

This change should force the release of version **5.0.0** instead of released version **4.14.0**.

dcavanagh:
@tiagoschenkel  I apologize for this error. I will work on releasing it as 5.0.0 instead

PR #915, opened 2018-07-12
Fix problem with applying options on resolve (fix #914)

### LINKED ISSUE #914: Container.resolve method doesn't use options
```ts
class Test { }

@injectable()
class Test2 extends Test { }

const container = new Container({skipBaseClassChecks: true});
container.resolve(Test2);
```

## Expected Behavior
Should resolve without error.

## Current Behavior
It throws "Error: Missing required @injectable annotation in: Test" even skipBaseClassChecks is true.

zandroid:
## Description

Fix #914 the problem with options usage in container.resolve(constructorFunction)

## Related Issue

Additionally this PR contains fix of tests in the first commit: test should use `.to.equal` for scalar values and when it needs to check strict equality (see http://www.chaijs.com/api/bdd/#equalval-msg). Especially before this change the test "Should be able to override options to child containers" was always green.

The second commit adds options inheritance for method `container.createChild` if no options are provided for child container. This commit improves functionality and has preparing for the final fix of the issue #914

## Types of changes

## Checklist:

### DIFF
--- a/test/planning/planner.test.ts
+++ b/test/planning/planner.test.ts
@@ -493,6 +493,16 @@ describe("Planner", () => {
         container.get(Test2);
     });
 
+    it("Should ignore checking base classes for @injectable on resolve when skipBaseClassChecks is set", () => {
+        class Test { }
+
+        @injectable()
+        class Test2 extends Test { }
+
+        const container = new Container({skipBaseClassChecks: true});
+        container.resolve(Test2);
+    });
+
     it("Should throw when an class has a missing @inject annotation", () => {
 
         interface Sword { }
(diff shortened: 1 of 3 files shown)


test/planning/planner.test.ts
```diff
@@ -493,6 +493,16 @@ describe("Planner", () => {
         container.get(Test2);
     });
 
+    it("Should ignore checking base classes for @injectable on resolve when skipBaseClassChecks is set", () => {
```
dcavanagh, line 496: You didn't talk about this in your PR comments. Was this just a test case that you saw was missing?

zandroid: This test case covers the issue #914

zandroid:
@dcavanagh can you release these changes? Because it requires to make monkey-patching of inversify v4.13.0 in my project to use it in such cases.

dcavanagh:
@zandroid Released this today. I apologize for the delay

zandroid:
@dcavanagh Thank you to notify me.

Issue #1115, opened 2019-08-01
Handling singletons in creation of large plans


vikram-cf:
We have a complex application with a huge dependency graph. On startup, when all controllers are initialized at once using inversify-express-utils (container.getAll for controllers) we noticed that the startup takes upward of 5 minutes, with more than 6 GB of memory taken for the plan graph. Most of our bindings are Singleton.

## Expected Behavior
container.getAll() should be fast

## Current Behavior
container.getAll() takes > 5 minutes, a huge amount of RAM

## Possible Solution
in planner.ts, we do not need to create childRequests for singleton Bindings multiple times, as there will be only instance of the object
Will be updating the issue with a pull request

## Steps to Reproduce (for bugs)
1. Application with a large dependency graph
2. container.get() an object which is dependent on the full dependency graph 

## Context

Speeds up the startup of our application, and reduces startup memory utilization from ~6.5 GB to 600MB

## Your Environment
node v11.13.0
inversify 5.0.1
inversify-express-utils 6.3.2
express 4.16.2

Private repo

# Stack trace
None

vikram-cf:
Related Pull request: https://github.com/inversify/InversifyJS/pull/1116

vikram-cf:
The fix seems to be having some issues with circular dependencies, trying to fix.

vikram-cf:
My approach did as caching during the planning phase seems to cause issues with dynamic bindings. Closing the PR

PR #1116, opened 2019-08-01
Added request caching for Singletons

vikram-cf:
## Description
For singleton bindings, ensuring that multiple Request objects are not created as there will be only one Instance finally.

## Related Issue
https://github.com/inversify/InversifyJS/issues/1115

## Motivation and Context
We have a complex application with a huge dependency graph. On startup, when all controllers are initialized at once using inversify-express-utils (container.getAll for controllers) we noticed that the startup takes upward of 5 minutes, with more than 6 GB of memory taken for the plan graph. Most of our bindings are Singleton. With this change, startup reduces to a few seconds, and about 600 MB memory

## How Has This Been Tested?
Tested my applications with the new changes, dependencies are being injected correctly. 
Also added/modified test cases as required

## Types of changes

## Checklist:

### DIFF
--- a/src/bindings/binding.ts
+++ b/src/bindings/binding.ts
@@ -42,6 +42,9 @@ class Binding<T> implements interfaces.Binding<T> {
     // On activation handler (invoked just before an instance is added to cache and injected)
     public onActivation: ((context: interfaces.Context, injectable: T) => T) | null;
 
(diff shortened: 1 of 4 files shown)


vikram-cf:
This change does not seem to work for Dynamic Bindings, hence closing.

Issue #471, opened 2017-01-05
Global activation hooks


captainjono:
Lets say i have a series of modules defined for an app. Each module adds certain features to the app. I know that modules will provide services that implement certain interfaces.

At runtime, i want to know when any of these services are first resolved from the container, either directly, or indirectly via a child dependency resolution.

Currently with middleware, i can only see a way to get access to an instance of the directly requested service. The only way i can get access to the instance of the child service is to use the OnActivation method thats created during the binding.

This is a cross cutting concern, i cant/or want to do this at registration time. It would be great if this could be supported. Currently im using a forked version of the repo to meet this requirement.

remojansen:
👍 thanks for this suggestion. Sorry for the late reply I was on holidays... I will try to do something about this and share my plans here soon...

fyodorvi:
I'm using this:
```
inversifyContainer.bind = <any>_.wrap(inversifyContainer.bind, (originalBind, ...rest: any[]) => {
	const binding = originalBind.apply(inversifyContainer, rest);
	binding._binding.onActivation = (context, instance) => {
		wireInstanceConfig(instance);
		return instance;
	};
	return binding;
});
```
Terribly dirty, I know, but that's the only way for now.

Moln:
I need this feature, too.

tonyhallett:
I will create a pull request for this tomorrow.

PR #1124, opened 2019-08-19
Pull request for Global activation hooks #471

tonyhallett:
## Description

Added onActivation method to the Container class.  This will add the handler to all the bindings that match the filter and optionally will do the same for ancestor containers.  Default behaviour is to wrap existing handlers and there is an option to overwrite them.

## Related Issue

inversify/monorepo#1040

## Motivation and Context

As per inversify/monorepo#1040, where there is common onActivation options are to store the syntax in array and iterate or to use middleware.  This enables doing so with a single method call.

## How Has This Been Tested?

TDD - checking that all/some of the Binding instances in the Container binding dictionary have had the onActivation handler present whether as is or wrapped with existing.  The filter too has been tested.

Tested on windows machine with current inversify test setup - mocha, sinon and chai.

Change does not affect other code.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x ] My code follows the code style of this project.
- [ x ] My change requires a change to the documentation.
- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.

I will happily update the docs if the pull request is accepted.

### DIFF
--- a/src/container/container.ts
+++ b/src/container/container.ts
@@ -269,6 +269,65 @@ class Container implements interfaces.Container {
         return tempContainer.get<T>(constructorFunction);
     }
 
+    public onActivation<T>(handler: interfaces.OnActivationHandler<T>, {
+        removeExisting = false,
+        filter = () => true,
+        setAncestors = false
+    }: interfaces.GlobalOnActivationOptions<T> = {}) {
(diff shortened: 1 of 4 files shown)


tonyhallett:
Just realised I was missing a verify call, I will change and create a new pull request shortly.

PR #1127, opened 2019-08-22
Binding.implementation ts correction

tonyhallett:
## Description

This is being picky but for correctness.  For the generic T to be correct the implementationType type has to change otherwise T will be incorrect for cache/onActivation when toConstructor has been used to set the Binding.

## Motivation and Context

It corrects the typescript - e.g no need for the BindingToSyntax to cast to any when setting the implementationType in the toConstructor method.

## How Has This Been Tested?

npm run test on Windows machine

## Types of changes

Yes if using the Binding.implementationType.
## Checklist:

No new tests required


notaphplover (approved):

src/bindings/binding.ts
```diff
@@ -15,8 +15,8 @@ class Binding<T> implements interfaces.Binding<T> {
     // A runtime identifier because at runtime we don't have interfaces
     public serviceIdentifier: interfaces.ServiceIdentifier<T>;
 
-    // The constructor of a class which must implement T
-    public implementationType: interfaces.Newable<T> | null;
+    // The constructor of a class
+    public implementationType: interfaces.Newable<any> | null;
```
notaphplover, line 19: I would use `unknown` instead

src/interfaces/interfaces.ts
```diff
@@ -51,7 +51,7 @@ namespace interfaces {
         dynamicValue: ((context: interfaces.Context) => T) | null;
         scope: BindingScope;
         type: BindingType;
-        implementationType: Newable<T> | null;
+        implementationType: Newable<any> | null;
```
notaphplover, line 54: I would use `unknown` instead

notaphplover (changes requested):
We are trying to avoid the use of any. As pointed previously, we shouldn't use the generic T in the implementationType property, but I think `unknown` is more appropiate even if additional casts may be added

tonyhallett:
It is not just the implementationType that is incorrect.  Closing as this needs more consideration.

Issue #1063, opened 2019-03-06
How to unbind with multi-injection?


akosyakov:
For:
```ts
container.bind<Weapon>("Weapon").to(Katana);
container.bind<Weapon>("Weapon").to(Shuriken);
```

I want to unbind only `Shuriken` from "Weapon". How can I do it?

dominic-simplan:
This might be a dulicate of #1035.

tonyhallett:
See #1035 for a solution

Issue #1035, opened 2019-01-18
Add ability to unbind a specific class


ghost:
I propose adding the ability to unbind a specific class from any previously bound interface.
If for example some project I depend on already binds the classes `Katana` and `Shuriken` to the interface `Weapon`, whereas I only want a multiinject to inject the `Katana` and other self-defined weapons and not shurikens, because they are just too sneaky.

## Expected Behavior
I unbind the shuriken from the weapon interface and bind a new weapon.
When using multiinject on the `Weapon` class only the `Katana` and my new weapon should be in the injected objects.

## Current Behavior
`unbind` unbinds all previously defined classes from the interface and only my new defined weapon would be in the multiinjected list.

## Possible Solution
The project I depend on already binds like this:
`bind(Katana).to(Weapon)`
`bind(Shuriken).to(Weapon)`
Add some new syntax to allow this:
`unbind(Shuriken).from(Weapon)`
`bind(Hammer).to(Weapon)`

dominic-simplan:
Did you find a workaround?

ghost:
As this was not too big of an issue for my use case, i simply ignored the problem.
A very hacky workaround would maybe be to unbind everything and just bind everything needed again. This may however add a lot of unnecessary dependencies.

tonyhallett:
You could use ContainerModule and unload, but this is better.  Note that it has not been carefully considered, but works for your example.  It relies upon the internal binding dictionary.

rebindContainer.ts
```
import {Container, interfaces, METADATA_KEY, BindingTypeEnum} from 'inversify'
import { getBindingDictionary} from 'inversify/dts/planning/planner'
const plannerExports=require('inversify/lib/planning/planner.js')
const gbd:typeof getBindingDictionary=plannerExports.getBindingDictionary;
type ServiceIdentifierOrArray=interfaces.ServiceIdentifier<any>|interfaces.ServiceIdentifier<any>[]
export class RebindContainer extends Container{
    private bd:interfaces.Lookup<interfaces.Binding<any>>
    constructor(options?:interfaces.ContainerOptions){
        super(options);
        this.bd=gbd(this);
    }
    private bindingMatchesTarget(binding:interfaces.Binding<any>,key:string | number | symbol,value:string){
        let found=false;
        if(binding.constraint&&binding.constraint.metaData){
            if(binding.constraint.metaData.key===key&&binding.constraint.metaData.value===value){
                found=true;
            }
        }
        return found;
    }
[...]

Issue #853, opened 2018-04-18
How to rebind named binding?


akosyakov:
See https://github.com/theia-ide/theia/pull/1664/files#r182040835

remojansen:
@akosyakov seems like we need to implement `rebindNamed` and `rebindTagged()`?

tonyhallett:
See #447 for a solution

tonyhallett:
I would happily do a pull request.
Should the container have the method ?
Should the binding syntax be extended ?
container.bind<IThing>(IThing).to(Thing) returns interface something like
```
interface UnbindRebind<T>{
unbind():void
rebind<T>:interfaces.BindingToSyntax<T>
rebindValue(newValue:any):void//or rebindTo/rebindToSelf
}
```

tonyhallett:
I have opted for extending the fluent syntax.  I should hopefully have a pull request tomorrow.

notaphplover:
Hey, since `inversify@7.2.0` you can now unbind a single binding. Refer to the [docs](https://inversify.io/docs/next/api/container/#example-unbind-a-binding-by-its-identifier) for more information.

Issue #447, opened 2016-12-14
Feature Request: Ability To Unbind @named Target


dac514:
Hi,

I am using the `@inject` syntax to do IoC. It's working as expected.

For unit tests I needed the ability to override the binding so that I could hotswap my class with a mock. Someone in https://github.com/inversify/InversifyJS/issues/143 suggested a clever way to this:

```typescript
kernel.unbind("HttpConfig")
kernel.bind("HttpConfig").to(customHttpConfig)
```

However, my issue is that the object I want to override is declared using this syntax:

```typescript
kernel.bind<ActiveDirectory>('Helper').to(ActiveDirectory).inSingletonScope().whenTargetNamed('ActiveDirectory');
```

Currently the only way I can override is:

```typescript
    kernel.unbind('Helper');
    kernel.bind<ActiveDirectory>('Helper').to(MockActiveDirectory).inSingletonScope().whenTargetNamed('ActiveDirectory');
```

The problem with this approach is that I am only interested in overriding the 'ActiveDirectory' class and not all Helpers. 

Help?

acopalipsis:
if I understand correctly, then I do it - 

```
// app-configuration.ts

export const app: Container = new Container();

app.bind<ISomeConfig>( Symbol.for( 'SomeConfig' ) ).to( { name: "APP_NAME" } );

// .spec.ts

import { app } from "../src/app-configuration.ts";

let spec: Container = new Container();
spec.parent = app;
spec.bind<ISomeConfig>( Symbol.for( 'SomeConfig' ) ).to( { name: "SPEC_NAME" } );

console.log( spec.get<ISomeConfig>( Symbol.for( 'SomeConfig' ) ).name ); // SPEC_NAME
```

remojansen:
Thanks for the prosal 👍  @acopalipsis solution is a good workaround. There are two workarounds at the moment:

- Use one kernel for test and one kernel for app
- Use kernel modules (kernel modules can load and unload groups of bindings)

Anyway this feature will be implemented in the future. It is likely to be released in the upcoming 3.1 release.

tonyhallett:
Here is a solution, although it has not been carefully considered but works for your example

rebindContainer.ts

```
import {Container, interfaces, METADATA_KEY, BindingTypeEnum} from 'inversify'
import { getBindingDictionary} from 'inversify/dts/planning/planner'
const plannerExports=require('inversify/lib/planning/planner.js')
[...]

PR #1131, opened 2019-08-22
Add unbind/rebind methods as terminals to the Binding fluent syntax

tonyhallett:
## Description

I have extended the fluent binding syntax so that it is now possible to :

```
container.bind("Warrior").to(Samurai);
const boundNinja = container.bind("Warrior").to(Ninja).InSingletonScope();
boundNinja.unbind();

or 
const boundAssassin=boundNinja.rebind().to(Assassin).inTransientScope()
```

The chain remains the same with the addition of BindingUnbindRebindSyntax methods available at each possible step.

**Important - small architectural change**

Before :
Each of the binding syntax classes constructed instances of the other binding syntax classes to facilitate the chaining.  Of course this worked but resulted in the binding being passed down the chain ( and for my case addition arguments ) and the tests were stubbing these private binding syntax instances.

After :
There is now a BindingSyntaxFactory which is constructed with the arguments required for the non pass through syntax classes.  The factory is used to create the binding syntax interfaces and it passes itself to the constructor of each binding syntax class, in addition to any arguments required by non pass through binding syntax classes.  In this way, none of the binding syntax classes depend on an implementation.  They now call through or return interfaces provided by the BindingSyntaxFactory

## Related Issue

#1035 
#1063 
#447 
#853 

## Motivation and Context

Container.unbind has its limitations, container modules can help in this regard but not ideal.

## How Has This Been Tested?

Old tests 
Some have been refactored - given that there is common pass through.
There were some missing expectations that have been added.
Now test against mock BindingSyntaxFactory

New tests - unbind/rebind syntax works / integration using the Container

## Types of changes

**Yes if people are constructing binding syntax classes - no if they are just using the fluent syntax**

**Future changes** - Adding the facility on the terminal BindingUnbindRebindSyntax to do the following:

```
const boundNinja = container.bind("Warrior").to(Ninja).inRequestScope().whenTargetNamed('Stealth');
const boundRequestScopeStealthAssassin = boundNinja.rebindTo(Assassin);
```

## Checklist:


NiklasRentzCAU:
This looks good already for the functionality, but there is one thing that could be a problem with this:
Usually if you want to unbind something (such as the Ninja in this example), that binding you do not want is part of another container that is loaded as well, but that you do not have direct access to. That way you will not have the variable (here: boundNinja) to call unbind() on and you still cannot get rid of that unwanted binding. Or is there another way that is not described in this PR?

tonyhallett:
@NiklasRentzCAU The child container has access to the parent. The parent could expose the BindingUnbindRebindSyntax objects as properties if desired.

jkhmtx:
@tonyhallett @dcavanagh Just cleaning up old PRs, please reopen if this was closed in error.

PR #1133, opened 2019-08-23
Fix 1-Hour Build Hangs on Appveyor

parisholley:
The Appveyor build environment has a newer version of Firefox that does not launch correctly with version 1.1 of the launcher. Without this fix, all PRs/commits will cause the repository to queue up for an hour each.


parisholley:
@tonyhallett you should merge this branch into your PR's so the Appveyor environment doesn't get queued up on commits

tonyhallett:
Do I need to change existing prs ?

parisholley:
@tonyhallett i would just so they show as green and incase you want to do more work :)

Issue #1143, opened 2019-09-02
Ability to inject dependencies for dynamicValue from same scope


thynson:
Make toDynamicValue binding able to injects its dependencies, to ensure its dependencies are 
resolved in **same scope**.

Motivation
-----------
I'm trying to implements a context logger.
```typescript
container.bind('ContextID')
  .toDynamicValue(()=> uuid())
  .inRequestScope();
container.bind('Logger')
  .toDynamicValue((context)=> {
    const targetName = (context.currentRequest.parentRequest.serviceIdentifier as Function).name;
    return new Logger(
      context.container.get('ContextID'),
      targetName
    )
  })
  .inTransientScope();

class FooComponent {

  @inject('Logger')
  loggerA: Logger;

}
class BarComponent {
  @inject('Logger')
  loggerB: Logger;
}
```
loggerA and loggerB cannot have same context id, even if they are resolved in same request, because `context.container.get('ContextID')` create a new request scope.

## Possible Solution
1. Allow inject depedencies into factory function, like nestjs does.
```typescript
container.bind('Logger').toDynamicValue({ inject: ['ContextID'], factory: (ctx)=> newLogger(ctx, ...)});
```
2. Make context itself be able resolve dependencies, in its own scope. 
```typescript
container.bind('Logger')
  .toDynamicValue((context)=> return new Logger(context.resolve('ContextID'), ...))
  .inRequestScope();
```

talha5389:
I also need this functionality. Right now, there does not seem to be a way to get a dependency already resolved in request scope in `toFactory`, `toDynamicValue` etc without resetting/recreating resolved object in request scope

dustinlacewell:
Oh god this hurts so bad. :(

notaphplover:
This feature has been implemented in the latest pre release and will be available with `inversify@7`: https://inversify.io/blog/announcing-inversify-7-0-0-alpha-4-prerelease

PR #1147, opened 2019-09-04
Root request scope

tonyhallett:
## Description

Added a new binding scope - inRootRequestScope()
This works similarly to request scope but applies to all container.get calls that occur during the 'root' container.get.  This allows scope to be shared across toDynamicValue
```
container.bind<Weapon>(TYPE.DynamicWeapon).toDynamicValue((context) => {
    return context.container.get<Weapon>(TYPE.Weapon);
});
```
This is facilitated by Context.parentContext, which is managed by Container._planAndResolve, and Context.rootRequestScope which is managed by the resolver.
## Related Issue

#1143 

## Motivation and Context

As soon as you use toDynamicValue you cannot resolve a previously resolved value unless it is singleton scoped.  With this code change there is now the concept of a root request which can be scoped to.

## How Has This Been Tested?

TDD.  Test that demonstrates what cannot be achieved with request scope that can now be achieved with root request scope.

## Types of changes

## Checklist:

As always, if you merge these changes I will happily update the documentation.


tonyhallett:
Parent context should be a stack.  I have a better abstraction that uses the ideas in this pull request that will add in a fresh pull request shortly.

PR #1307, opened 2021-04-14
Replace any

PodaruDragos:
this is based on the other PR changing to eslint.
this is the branch we talked about. might need a lot of work tho


PodaruDragos:
I'm closing this since it will be very hard to do it in this way.
me and @notaphplover will open another PR on another branch and we will take it step by step.

1. Update typescript with more restrictive options
2. Solve all the compiler errors
3. Create small commits which keeps the code with 0 compiler errors
4. Add eslint

Issue #928, opened 2018-07-27
Construtor injection is broken for TypeScript and es6


AlexTugarev:
The constructor injection breaks after switching TypeScript's target from es5 to es6 in a specific case. Consider a constructor of a base class with an optional dependency and a default value, and the subclass' constructor to have more dependencies then the super constructor.

## Expected Behavior
Constructor injection should have the same behavior regardless of target in TS (es5/es6), and dependencies are injected as specified.

## Current Behavior
Constructor injection fails after switching to es6 target in that particular case, see repro. Only `n` dependencies are injected, where `n` is the number of super constructor arguments/dependencies. 

## Steps to Reproduce (for bugs)
This issue is reproducible with the following minimal example:
```typescript
import { injectable, inject, Container, optional } from "inversify";
import "reflect-metadata";

// some dependencies
@injectable() export class DepA {a=1}
@injectable() export class DepB {b=1}
@injectable() export class DepC {c=1}

@injectable() export abstract class AbstractCls {
    constructor(@inject(DepA) a: DepA, @inject(DepB) @optional() b: DepB = {b: 0}) {
        console.log(`Expect {"a":1}, Actual: ${JSON.stringify(a)}`);
        console.log(`Expect {"b":1}, Actual: ${JSON.stringify(b)}`);
    }
}

@injectable() export class Cls extends AbstractCls {
    constructor(@inject(DepC) c: DepC, @inject(DepB) @optional() b: DepB = { b: 0}, @inject(DepA) a: DepA) {
        super(a, b);
    }
}

const container = new Container();
[DepA, DepB, DepC, Cls].forEach(i => container.bind(i).toSelf().inSingletonScope());
container.get(Cls);

// TS > es5
// [start] Expect {"a":1}, Actual: {"a":1}
// [start] Expect {"b":1}, Actual: {"b":1}

// TS > es6
// [start] Expect {"a":1}, Actual: undefined
// [start] Expect {"b":1}, Actual: {"b":0}
```

#### tsconfig.json
```json
{
  "compilerOptions": {
    "target": "es6",
    "module": "commonjs",
    "lib": ["es6", "dom"],
    "sourceMap": true,
    "outDir": "dist",
    "strict": true,
    "moduleResolution": "node",
    "types": ["reflect-metadata"],
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true
  }
}
```

## Context
[...]

PR #1308, opened 2021-04-14
Fix injection of optional properties

notaphplover:
## Description

This PR updates the planner to get the right bindings when a constructor has optional parameters and adds ES6 ci test jobs to ensure the library works fine when ES6 is the Typescript compilation target.

## Related Issue

#928 

## Motivation and Context

A constructor with injected rest parameters won't be injected with those rest parameters injection. Typescript users who targets ECMAScript 6 could have issues because of this.

## How Has This Been Tested?

A test case have been added.
New ci test jobs have been added.

## Types of changes

## Checklist:


test/bugs/issue_543.test.ts
```diff
@@ -19,21 +19,21 @@ describe("Issue 543", () => {
 
         @injectable()
         class Child2 {
-            public circ: Circular;
+            public circ: unknown;
```
PodaruDragos: do we need unknown here ?

notaphplover: I think yes. If you try to run the tests targeting ES6 and you do not change typings, you get the folloing error:

```
ReferenceError: Cannot access 'Circular' before initialization
      at Context.<anonymous> (test/bugs/issue_543.test.ts:24:46)
      at processImmediate (internal/timers.js:456:21)
```

I think we can change the type because at the end it's sugar syntax. What we really want to check is we are able to detect a circular dependency but maybe I'm missing something

PodaruDragos (approved):
this looks fine to me, but maybe we can find a way to actually keep types as types there in tests.
I guess we can do that later

notaphplover:
> this looks fine to me, but maybe we can find a way to actually keep types as types there in tests.
> I guess we can do that later

Yeah, I'd love to find a way. I've been compiling the code targeting es6 modules and targeting es5:

The following code fragment:

```Typescript
@injectable()
class Child2 {
    public circ: Circular;
    public constructor(
        @inject(TYPE.Circular) circ: Circular
    ) {
        this.circ = circ;
    }
}
...
@injectable()
class Circular {
    public irrelevant: unknown;
    public child: unknown;
    public constructor(
        @inject(TYPE.Irrelevant) irrelevant: Irrelevant,
        @inject(TYPE.Child) child: Child
    ) {
        this.irrelevant = irrelevant;
        this.child = child;
    }
}
```

Is compiled into this fragment: (ES5)

```js
var Child2 = (function () {
    function Child2(circ) {
        this.circ = circ;
    }
    Child2 = __decorate([
        inversify_1.injectable(),
        __param(0, inversify_1.inject(TYPE.Circular)),
        __metadata("design:paramtypes", [Circular])
    ], Child2);
    return Child2;
}());
...
var Circular = (function () {
    function Circular(irrelevant, child) {
        this.irrelevant = irrelevant;
        this.child = child;
    }
    Circular = __decorate([
        __param(0, inversify_1.inject(TYPE.Irrelevant)),
        __param(1, inversify_1.inject(TYPE.Child)),
        __metadata("design:paramtypes", [Irrelevant, Child])
    ], Circular);
    return Circular;
}());
```

So we are "fine" thanks to the `var` scope (even if metadata is not being decorated properly).

Let's see the same compiled fragment on es6:

```js
let Child2 = class Child2 {
    constructor(circ) {
        this.circ = circ;
    }
};
Child2 = __decorate([
    inversify_1.injectable(),
    __param(0, inversify_1.inject(TYPE.Circular)),
    __metadata("design:paramtypes", [Circular])
], Child2);
...
let Circular = class Circular {
    constructor(irrelevant, child) {
        this.irrelevant = irrelevant;
        this.child = child;
    }
};
Circular = __decorate([
    __param(0, inversify_1.inject(TYPE.Irrelevant)),
    __param(1, inversify_1.inject(TYPE.Child)),
    __metadata("design:paramtypes", [Irrelevant, Child])
], Circular);
```

`let` scope is being used instead, so it won't work.

Since we need to emit class metadata and one class must be the first one (splitting classes in multiple files wouln't work the problem since we would be creating a circular dependency), I wonder if we can solve this :( without the `unknown` workaround.

test/bugs/issue_543.test.ts
dcavanagh: I was working on this yesterday and rewrote the tests I will try to add just that part

.vscode/launch.json
```diff
@@ -10,21 +10,18 @@
...
             ],
-            "sourceMaps": true,
-            "outFiles": [
-                "${workspaceRoot}/test",
-                "${workspaceRoot}/src"
-            ],
-            "internalConsoleOptions": "openOnSessionStart"
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen",
+            "protocol": "inspector"
```
notaphplover, line 24: :heart:

notaphplover:
> this looks fine to me, but maybe we can find a way to actually keep types as types there in tests.
> I guess we can do that later

At the @dcavanagh was able to find a way though interfaces :smiley:

dcavanagh (approved):
Fixes #928

Issue #1271, opened 2020-12-13
Looks like incorrect behavior is in _validateActiveBindingCount


tryvols:
https://github.com/inversify/InversifyJS/blob/65596eaa2915787d14f852992539a957acc23af0/src/planning/planner.ts#L116

```typescript
      switch (bindings.length) {
         // ...
         case BindingCount.OnlyOneBindingAvailable:
            if (!target.isArray()) {
                return bindings;
            }
        case BindingCount.MultipleBindingsAvailable:
        default:
            if (!target.isArray()) {
                // ...
            } else {
                return bindings;
            }
      }
```

If `bindings.length === BindingCount.OnlyOneBindingAvailable` and:
- target isn't array - function returns `bindings`
- target is array - we fall through to default condition, then in else block function returns `bindings`

So, in case **BindingCount.OnlyOneBindingAvailable** - function **always returns `bindings`**.

Looks like it's not correct behavior.

notaphplover:
Hi @tryvols I'll have a look at it :)

notaphplover:
Hi @tryvols after having a look we could simplify the code, but the behavior is correct

### What means that target is Array

If look at `Target` implementation:

```Typescript
public isArray(): boolean {
    return this.hasTag(METADATA_KEY.MULTI_INJECT_TAG);
}
```

It means this service is being requested in a multi inject context. In this case  we should return the array of bindings even if this is a multiInject context and we only got one binding. In the following code:

```Typescript
import { Container, inject, injectable } from 'inversify';

@injectable()
Class A {}

@injectable()
Class B {
  @multiInject('a')
  constructor(
      private readonly a: A[]
  )
}

const container = new Container();
container.bind('a').to(A);
container.bind('b').to(B);

const b = container.get<B>('b');
```

We expect b to be injected with an array of one `A` element even if it's a multi inject context

For the same reason `container.getAll('a')` should return an array of one `A` element, even if we are in a `@multiInject` context.

We have plans to refactor this

Issue #1222, opened 2020-06-02
Annotating dependencies with inversify.inject() and inversify.decorate() produces a TypeScript type error


baronswindle:
My project is using TypeScript for type checking and Babel to compile the TypeScript source to JavaScript. Because the current TC39 proposal for decorators does not current include support for parameter decorators, we're using the `inversify.decorate()` method to annotate our classes. However, when we use the `inversify.inject()` decorator, we're getting a type error.

```
Argument of type '(target: any, targetKey: string, index?: number | undefined) => void' is not assignable to parameter of type 'ClassDecorator | ParameterDecorator | MethodDecorator'.
  Type '(target: any, targetKey: string, index?: number | undefined) => void' is not assignable to type 'ClassDecorator'.
```

Root cause seems to be that the return type of `inversify.inject()` is not compatible with the `ParameterDecorator` type from TypeScript's `lib.es5.d.ts`. The `targetKey` argument can be a `string` or `symbol` in `ParameterDecorator`, but it's defined as `string` only in the return type of `inversify.inject()`.

## Expected Behavior
The following lines of code SHOULD NOT generate a 
```ts
inversify.decorate(inversify.inject(TYPES.KATANA), Ninja, 0);
inversify.decorate(inversify.inject(TYPES.SHURIKEN), Ninja, 1);
```

## Current Behavior
The lines of code above DO produce a type error.

## Steps to Reproduce (for bugs)
I've pushed up a pretty minimal example to reproduce the type error.
https://github.com/baronswindle/inversify-bug/tree/6149a7694d124aa84796aba463615142ec5565bd

Once you've cloned the repo, run `npm install` followed by `npx tsc`.

monteki:
Same here. Any solutions?

vzkhrv:
> ```
> Argument of type '(target: any, targetKey: string, index?: number | undefined) => void' is not assignable to parameter of type 'ClassDecorator | ParameterDecorator | MethodDecorator'.
>   Type '(target: any, targetKey: string, index?: number | undefined) => void' is not assignable to type 'ClassDecorator'.
> ```

Same here. Any solutions?

sonewman:
I’m no wizard, but I think you should be using `injectable` and not `inject`

valjusha:
@baronswindle Where did you end up?

baronswindle:
[...]

PR #1132, opened 2019-08-22
Promise (Lazy) Support + Global onActivation/onDeactivation

parisholley:
This is a cleaner refactor which combines #1074 and #1086 to allow for users to opt-in to lazy loaded dependencies by using an asynchronous API.

For anyone wanting to play around with this, you can use the NPM release:

`@parisholley/inversify-async:1.1.7`

## Description

New container methods for pulling objects out of the container in an asynchronous fashion while respecting container semantics and error if trying to mix synchronous/asynchronous dependencies.

**Implementation**
```javascript
container.bind<Type>('name').toDynamicValue(async () => {
  return await myAsyncFunction();
});
container.bind<string>('resolve').toDynamicValue(() => Promise.resolve('foobar'));

const value = await container.getAsync<Type>('name'); // result of myAsyncFunction
const value = await container.get<Type>('name'); // throws an exception
const value = await container.getAsync<string>('resolve'); // foobar
````
**Global onActivation**

```javascript
container.onActivation('asyncIdentifier', (context, injectable) => {
  return Promise.resolve('newObject'); // returning a promise enforces use of getAsync(), even if future binds are not async
});
container.bind('asyncIdentifier').toConstantValue('foobar');

await container.getAsync('asyncIdentifier');

container.onActivation('syncIdentifier', (context, injectable) => {
  return 'newObject'; // no promise = use legacy API
});

container.get('syncIdentifier');
````

**Binding onActivation**

```javascript
container.bind('asyncIdentifier').toConstantValue('foobar').onActivation((context, injectable) => {
  return Promise.resolve('newObject'); // returning a promise enforces use of getAsync()
});

await container.getAsync('asyncIdentifier');

container.bind('asyncIdentifier').toConstantValue('foobar').onActivation((context, injectable) => {
  return 'newObject'; // no promise = use legacy API
});

container.get('syncIdentifier');
````

**Parent/Child onActivation**

```javascript
container.onActivation('asyncIdentifier', (context, injectable) => {
  return Promise.resolve('newObject'); // returning a promise enforces use of getAsync()
});

const child = container.createChild();
child.bind('asyncIdentifier').toConstantValue('foobar');

await container.getAsync('asyncIdentifier');
````

**Module onActivation**

```javascript
container.bind('asyncIdentifier').toConstantValue('foobar');

const TestContainerModule = new ContainerModule((bind, unbind, isBound, rebind, onActivation) => {
  onActivation('asyncIdentifier', async (context, service) => {

 }));
});
````

[...]

PR #1311, opened 2021-04-18
Update typescript compiler options

notaphplover:
## Description

This PR aims to update Typescript version and add strict compiler options

## Related Issue

#1222, #1271

## Motivation and Context

We should try to use [Typescript strict options](https://www.typescriptlang.org/tsconfig/#strict) in our project. This would help us to detect some typing issues such as #1222, which would have been detected.

## How Has This Been Tested?

- All tests passes successfully

## Types of changes

## Checklist:


src/planning/planner.ts
```diff
@@ -114,10 +114,7 @@ function _validateActiveBindingCount(
             }
 
         case BindingCount.OnlyOneBindingAvailable:
-            if (!target.isArray()) {
-                return bindings;
-            }
-
+            return bindings;
```
PodaruDragos: we still need to throw something here

notaphplover: Yeah, we need to investigate that case

notaphplover: @PodaruDragos I investigated this case in an issue, I think this code is right, we should allways return the bindings, see [this comment](https://github.com/inversify/InversifyJS/issues/1271#issuecomment-822444015) for more information

package.json
```diff
@@ -71,7 +71,7 @@
     "ts-node": "9.1.1",
     "tsify": "^5.0.2",
     "tslint": "6.1.3",
-    "typescript": "4.1.3",
+    "typescript": "^4.2.4",
```
notaphplover: Typescript creates false minor versions, we should ensure we are using the right version
```suggestion
    "typescript": "4.2.4",
```

PodaruDragos: agree

PodaruDragos:
@tonyhallett @notaphplover I really want to get this merged eventually. Things at this moment should be correct for building but tests are failing and I really don't know what's happening there. Maybe some of you can have a second opinion on this.

tonyhallett:
@PodaruDragos I am busy finishing a project at the moment.  Might get the chance to take a look tomorrow or Thursday

PodaruDragos:
> @PodaruDragos I am busy finishing a project at the moment. Might get the chance to take a look tomorrow or Thursday

no worries, take your time

PodaruDragos:
@tonyhallett @notaphplover I've kept this branch up to date. There are 2 issues when running `test` and 1 issue when running `test:browser` and `build`.

Can you also take a look at what's going on there maybe you have a clearer picture on how to solve those issues.
Then I think we can merge this safely and focus on the `binding-abstractions` with stricter `typescript`.

tonyhallett:
Looking at it now

tonyhallett:
image

Rectified with change to

```
    public toAutoNamedFactory<T2>(serviceIdentifier: interfaces.ServiceIdentifier<T2>): BindingWhenOnSyntax<T> {
        this._binding.type = BindingTypeEnum.Factory;
        this._binding.factory = (context) => {
            return (named: unknown) => context.container.getNamed<T2>(serviceIdentifier, named as string);
        };
        return new BindingWhenOnSyntax<T>(this._binding);
    }
```

tonyhallett:
I would also say that the [pull](https://github.com/inversify/InversifyJS/pull/1370) by @LeonardSSH  has introduced an issue in the typings.

```
export type Factory<T, U extends unknown[] = unknown[]> = (...args: U) => (((...args: U) => T) | T);
```
When
```
 (...args: U) => (...args: U) => T
```
Why should the second arguments be the same as the first ?

This error can be seen in https://github.com/inversify/InversifyJS/blob/731aa753c2b6746df5ce69b25c1efb11d3c03188/test/inversify.test.ts#L900



```
        container.bind<interfaces.Factory<Engine>>("Factory<Engine>").toFactory<Engine>((context: interfaces.Context) =>
            (theNamed: string) => (displacement: number) => {
                const theEngine = context.container.getNamed<Engine>("Engine", theNamed);
                theEngine.displacement = displacement;
                return theEngine;
            });
```

Also note that because of unknown you have to provide the second generic to 'toFactory' ( or have unknown parameters )

tonyhallett:
After clearing the issue identified by ci build I then noticed in my error list that there were 50ish other errors with many that are not pertaining to 'update typescript compiler options' but are a result of the merge and the fix leaving the code in this present state.

Please resolve resolver.test.ts and get Factory/FactoryCreator sorted.
Note that Provider/ProviderCreator have not changed. 
https://github.com/inversify/InversifyJS/blob/731aa753c2b6746df5ce69b25c1efb11d3c03188/src/interfaces/interfaces.ts#L85

Then I will help look at the decorator issues and any strange ci browser test issue that may present itself.

PodaruDragos:
@tonyhallett checked the branch right now, i don't have any issues but the one already mentioned

PodaruDragos:
I also formatted the code to have 2 spaces all over the place.. The 4 and 2 spaces randomly was really freaking me out.
This was also agreed with Dan when we eventually do finish this and move to `eslint` that we should use 2 spaces for formatting.

@tonyhallett  hope that didn't cause you any discomfort.

tonyhallett:
> checked the branch right now, i don't have any issues but the one already mentioned

I will check to see what stupid mistake I made - apologies in advance

tonyhallett:
Compare this pull request https://github.com/inversify/InversifyJS/blob/3aa77dec705537eb97df769ab5a1368511a90c98/test/resolution/resolver.test.ts#L1134

With master
https://github.com/inversify/InversifyJS/blob/731aa753c2b6746df5ce69b25c1efb11d3c03188/test/resolution/resolver.test.ts#L1076

There is no katana on Ninja

image


This commit https://github.com/inversify/InversifyJS/commit/27dd0429f573134414ea50721b772837d3e08433 looks to me to be the culprit ( a previous merge to the one I said )

PodaruDragos:
You are right. I'll go over each file from tests again ( copy-pasting from master and changing it to work with typescript )
I'll flag the errors that are still there. Maybe then you can help sort them out.

Thanks for taking the time.

PodaruDragos:
@tonyhallett hello I've copy pasted tests from master to this branch and fixing the typescript errors.
Now the left ones are the ones that we know about from my previous comment.

tonyhallett:
There are now 12 errors. 6 errors unrelated to decorators are easily fixed.   

Please can you look at my comments https://github.com/inversify/InversifyJS/pull/1311#issuecomment-905480952 https://github.com/inversify/InversifyJS/pull/1311#issuecomment-905529053.

Hopefully I will get the opportunity to look at the decorators later today if not tomorrow.

tonyhallett:
These are the changes that I have made locally

add esModuleInterop:true to test/tsconfig.json
change imports
import * as Proxy from "harmony-proxy";  -> import Proxy from "harmony-proxy";
in binding_on_syntax.test.ts,  resolver.test.ts and proxies.test.ts
import * as now from "performance-now"; -> import now from "performance-now";
in performance.test.ts and request_scope.test.ts

decorator changes

inject_base.ts
change from `TypedPropertyDescriptor<unknown>` to `TypedPropertyDescriptor<any>`

decorator_utils.ts
change from `TypedPropertyDescriptor<unknown>` to `TypedPropertyDescriptor<any>`
change decorate function parameter decorator from `decorator: (ClassDecorator | ParameterDecorator | MethodDecorator)` to `(DecoratorTarget | ParameterDecorator | MethodDecorator)`



post_construct.test
remove - as unknown as ClassDecorator

delete performance-now declaration from test/globals.d.ts

PodaruDragos:
I just realized that we can drop `harmony-proxy` and `performace-now` since javascript has them and we are also using typescript.
Maybe when the project started there was no support for this.

Let me try removing them  and see what's what

tonyhallett:
Ignore what I said about decorate - looking at this further now

PodaruDragos:
> inject_base.ts
> change from `TypedPropertyDescriptor<unknown>` to `TypedPropertyDescriptor<any>`

I still think we should now use any here. What are all the types we can use here ? ( maybe make them a union of all possible types )

PodaruDragos:
pushed a version where I remove old packages and also the `global.d.ts` file since it was not needed

tonyhallett:
> I still think we should now use any here.

I assume that you mean "not use".

The reason why it is necessary is that typescript is able to determine for properties the specific type - e.g `TypedPropertyDescriptor<Warrior>`.

tonyhallett:
> Ignore what I said about decorate - looking at this further now

Continue as before IMO

tonyhallett:
You could do this and have only the one any

```
function createTaggedDecorator(
  metadata: interfaces.MetadataOrMetadataArray,
) {
  return <T>(
    target: DecoratorTarget,
    targetKey?: string | symbol,
    indexOrPropertyDescriptor?: number | TypedPropertyDescriptor<T>,
  ) => {
    if (typeof indexOrPropertyDescriptor === "number") {
      tagParameter(target, targetKey, indexOrPropertyDescriptor, metadata);
    } else {
      tagProperty(target, targetKey as string | symbol, metadata);
      }
  };
}
```

tonyhallett:
Ultimately the type is only there to signify that there are two usages of the decorator.  The `TypedPropertyDescriptor<T>` is never used

```
export function injectBase(metadataKey: string) {
  return (serviceIdentifier: ServiceIdentifierOrFunc) => {
    return (
      target: DecoratorTarget,
      targetKey?: string | symbol,
      indexOrPropertyDescriptor?: number | TypedPropertyDescriptor<any>,
    ) => {
      if (serviceIdentifier === undefined) {
        const className = typeof target === "function" ? target.name : target.constructor.name;

        throw new Error(UNDEFINED_INJECT_ANNOTATION(className));
      }
      return createTaggedDecorator(
        new Metadata(metadataKey, serviceIdentifier)
      )(target, targetKey, indexOrPropertyDescriptor);
    };
  }
}
```

```
function createTaggedDecorator(
  metadata: interfaces.MetadataOrMetadataArray,
) {
  return <T>(
    target: DecoratorTarget,
    targetKey?: string | symbol,
    indexOrPropertyDescriptor?: number | TypedPropertyDescriptor<T>,
  ) => {
    if (typeof indexOrPropertyDescriptor === "number") {
      tagParameter(target, targetKey, indexOrPropertyDescriptor, metadata);
    } else {
      //**************************** TypedPropertyDescriptor<T> not used
      tagProperty(target, targetKey as string | symbol, metadata);
      }
  };
}
```

tonyhallett:
> What are all the types we can use here ? ( maybe make them a union of all possible types )

Get a union of all the setters ?

I don't think it is possible to distinguish between field and property and if you could it is the same issue

`Type TypedPropertyDescriptor<Thing> is not assignable to TypedPropertyDescriptor<Thing | OtherThing>`

PodaruDragos:
in `decorator_utils.test`

`  @inject("Thing")`

`
Argument of type 'TypedPropertyDescriptor<Thing>' is not assignable to parameter of type 'number | TypedPropertyDescriptor<unknown> | undefined'.
  Type 'TypedPropertyDescriptor<Thing>' is not assignable to type 'TypedPropertyDescriptor<unknown>'.
    Types of property 'set' are incompatible.
      Type '((value: Thing) => void) | undefined' is not assignable to type '((value: unknown) => void) | undefined'.
        Type '(value: Thing) => void' is not assignable to type '(value: unknown) => void'.
          Types of parameters 'value' and 'value' are incompatible.
            Type 'unknown' is not assignable to type 'Thing'.ts(2345)
 `

PodaruDragos:
@tonyhallett  unrelated to this

What i think the Metadata file should look like is

```
export type MetadataValue<T> = string | number | symbol | boolean | ServiceIdentifierOrFunc<T>;

class Metadata<T = MetadataValue<unknown>> implements interfaces.Metadata {

  public key: string | number | symbol;
  public value: MetadataValue<T>;

  public constructor(
    key: string | number | symbol,
    value: MetadataValue<T>
  ) {
    this.key = key;
    this.value = value;
  }

  public toString() {
    if (this.key === METADATA_KEY.NAMED_TAG) {
      return `named: ${this.value.toString()} `;
    } else {
      return `tagged: { key:${this.key.toString()}, value: ${String(this.value)} }`;
    }
  }
}

export { Metadata };
```
This way we cover all values we need and if we do need another one 

```
// Used to add custom metadata which is used to resolve metadata-based contextual bindings.
function tagged<T>(metadataKey: string | number | symbol, metadataValue: MetadataValue<T>) {
```

I'll commit this, please let me know what you think

PodaruDragos:
@tonyhallett @notaphplover  from my point of view the work here is done. everything passes locally.

I think the CI is mentally since `perfornance.now` exists in `node` from version 8 something.
Could you guys take a look into this ?

tonyhallett:
> perfornance.now exists in node from version 8 something.

It does but you need to require it
https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_measurement_apis
`const { PerformanceObserver, performance } = require('perf_hooks');`

and include `@types/node`

Because the root tsconfig has
```
{
  "compilerOptions": {

    "lib": [
      "esnext",
      "dom"
    ],
```
there are no typescript errors.

The reason why it works locally and not on the server is that it is global version dependent.

Global in latest https://nodejs.org/dist/latest-v16.x/docs/api/globals.html#globals_performance
Not global in v14 https://nodejs.org/dist/latest-v14.x/docs/api/globals.html

We could change the test directory so that there is a tsconfig for browser tests and tsconfig for node tests.

tonyhallett:
I will add a quick review today.

src/annotation/decorator_utils.ts
```diff
@@ -126,17 +128,17 @@ function _param(paramIndex: number, decorator: ParameterDecorator) {
...
-    parameterIndexOrProperty?: number | string): void {
-
-    if (typeof parameterIndexOrProperty === "number") {
-        _decorate([_param(parameterIndexOrProperty, decorator as ParameterDecorator)], target);
-    } else if (typeof parameterIndexOrProperty === "string") {
-        Reflect.decorate([decorator as MethodDecorator], target, parameterIndexOrProperty);
-    } else {
-        _decorate([decorator as ClassDecorator], target);
-    }
+  decorator: (ClassDecorator | ParameterDecorator | MethodDecorator),
```
tonyhallett: (DecoratorTarget | ParameterDecorator | MethodDecorator) I think that there was confusion from the comment I made.  Then casts can go.

src/interfaces/interfaces.ts
```diff
@@ -1,362 +1,363 @@
...
+    onActivation: BindingActivation<TActivated> | null;
+    onDeactivation: BindingDeactivation<TActivated> | null;
+    cache: null | TActivated | Promise<TActivated>;
+  }
+
+  export type Factory<T> = (...args: any[]) => (((...args: any[]) => T) | T);
+
+  export type FactoryCreator<T> = (context: Context) => Factory<T>;
+
+  export type FactoryTypeFunction<T = unknown> = (context: interfaces.Context) => any;
```
tonyhallett: T is not used

src/interfaces/interfaces.ts
```diff
@@ -1,362 +1,363 @@
...
+
+  export interface FactoryDetails<T = unknown> {
+    factoryType: FactoryType,
+    factory: FactoryTypeFunction<T> | null
```
tonyhallett: remove the generic if not used

src/interfaces/interfaces.ts
```diff
@@ -1,362 +1,363 @@
...
+    factory: FactoryCreator<unknown> | null;
+    provider: ProviderCreator<unknown> | null;
+    onActivation: BindingActivation<TActivated> | null;
+    onDeactivation: BindingDeactivation<TActivated> | null;
+    cache: null | TActivated | Promise<TActivated>;
+  }
+
+  export type Factory<T> = (...args: any[]) => (((...args: any[]) => T) | T);
+
+  export type FactoryCreator<T> = (context: Context) => Factory<T>;
```
tonyhallett: The original was acceptable apart from the second U
Of course would be a breaking change defaulting U to unknown[].
```
export type Factory<T, U extends unknown[] = unknown[]> = (...args: U) => (((...args: *************) => T) | T);
export type FactoryCreator<T, U extends unknown[] = unknown[]> = (context: Context) => Factory<T, U>;

```

I also wonder why Provider/ProviderCreator is not typed the same

src/interfaces/interfaces.ts
```diff
@@ -1,8 +1,9 @@
 import { FactoryType } from "../utils/factory_type";
 
 namespace interfaces {
-  export type DynamicValue<T> = (context: interfaces.Context) => T | Promise<T>
-  export type ContainerResolution<T> = T | Promise<T> | (T | Promise<T>)[]
+  export type Unpacked<T> = T extends (infer U)[] ? U : T;
```
tonyhallett: This is not used

test/annotation/decorator_utils.test.ts
```diff
@@ -1,56 +1,58 @@
...
 
     @injectable()
     class Root {
-      public thingyType:string;
-      @multipleMetadataDecorator("Key1Value","Key2Value")
-      @inject("Thing")
-      set thingy(thingy:Thing) {
+      public thingyType!: string;
+      @multipleMetadataDecorator("Key1Value", "Key2Value")
+      @inject<Thing>("Thing")
```
tonyhallett, line 54: I don't think that a dev should need to supply the generic to satisfy your need to avoid `TypedPropertyDescriptor<any>`

src/annotation/lazy_service_identifier.ts
```diff
@@ -1,8 +1,8 @@
 import { interfaces } from "../interfaces/interfaces";
 
-export type ServiceIdentifierOrFunc = interfaces.ServiceIdentifier<any> | LazyServiceIdentifer;
+export type ServiceIdentifierOrFunc<T> = interfaces.ServiceIdentifier<T> | LazyServiceIdentifer;
```
tonyhallett: Pass T to LazyServiceIdentifier.

src/resolution/instantiation.ts
```diff
@@ -5,148 +5,150 @@ import { interfaces } from "../interfaces/interfaces";
...
-        const data: Metadata = Reflect.getMetadata(METADATA_KEY.POST_CONSTRUCT, constr);
-        try {
-            return (instance as T & Record<string, () => void>)[data.value]();
-        } catch (e) {
-            throw new Error(POST_CONSTRUCT_ERROR(constr.name, e.message));
-        }
+  if (Reflect.hasMetadata(METADATA_KEY.POST_CONSTRUCT, constr)) {
+    const data: Metadata = Reflect.getMetadata(METADATA_KEY.POST_CONSTRUCT, constr);
+    try {
+      return (instance as T & Record<string, () => void>)[(data.value as number)]?.();
```
tonyhallett: why have you cast as number ?

src/planning/metadata.ts
```diff
@@ -1,12 +1,18 @@
 import * as METADATA_KEY from "../constants/metadata_keys";
 import { interfaces } from "../interfaces/interfaces";
+import { ServiceIdentifierOrFunc } from "../annotation/lazy_service_identifier";
 
-class Metadata implements interfaces.Metadata {
+export type MetadataValue<T> = string | number | symbol | boolean | ServiceIdentifierOrFunc<T>;
+
+class Metadata<T = MetadataValue<unknown>> implements interfaces.Metadata {
```
tonyhallett: What are we gaining from this ?

Metadata can have any type of value.

binding_when_syntax whenTargetTagged ->
https://github.com/inversify/InversifyJS/blob/731aa753c2b6746df5ce69b25c1efb11d3c03188/src/syntax/constraint_helpers.ts#L25

The value T is not necessarily associated with ServiceIdentifierOrFunc.
So this really does not make sense `function tagged<T>(metadataKey: string | number | symbol, metadataValue: MetadataValue<T>) {`

Why not just change the type of value to unknown ? ( which would require `named: ${String(this.value)}`
Then change Target accordingly but then T is used only for ServiceIdentifier.  If you then change 
`export type ServiceIdentifier<T> = (string | symbol | Newable<T> | Abstract<T>);`
to `export type ServiceIdentifier<T = unknown> = (string | symbol | Newable<T> | Abstract<T>);`
then is there any point of T.  Inversify never creates a typed Target and it is not exported.

src/planning/target.ts
```diff
@@ -5,110 +5,112 @@ import { getSymbolDescription } from "../utils/serialization";
...
-        // target has metadata
-        if (metadataItem !== null) {
-            this.metadata.push(metadataItem);
-        }
-
+class Target<T = unknown> implements interfaces.Target {
+
+  public id: number;
+  public type: interfaces.TargetType;
+  public serviceIdentifier: interfaces.ServiceIdentifier<any>;
```
tonyhallett: `ServiceIdentifier<T>` but given comment from before T imo is not necessary.

tonyhallett (commented):
I cannot dedicate anymore time looking at this in finer detail.
Due to formatting and copy paste it is difficult to see the changes you have made.

PodaruDragos:
@tonyhallett thanks for reviewing this, I'll try to go over all the comments and address the feedback. I'll tag you once I'll do this.

tonyhallett:
Please see amended comment on performance.now https://github.com/inversify/InversifyJS/pull/1311#issuecomment-907092941

src/resolution/instantiation.ts
PodaruDragos: because value is unknown

src/planning/metadata.ts
PodaruDragos: modified this so now `Metadata` has a value of `unknown` and aplied changes where neded

src/interfaces/interfaces.ts
PodaruDragos: I can't seem to get this to work with `unknown`. Can you maybe just push into this branch what are you refering to here ?

src/annotation/decorator_utils.ts
PodaruDragos: if we get rid of the casts then what "arch" type should we use there ?

src/resolution/instantiation.ts
tonyhallett: And what type is data.value

src/annotation/decorator_utils.ts
tonyhallett: I wasn't referring to the casts in the method.  The cast in the method to ClassDecorator to DecoratorTarget though

src/interfaces/interfaces.ts
tonyhallett: As before except U should not be used where *******

src/resolution/instantiation.ts
PodaruDragos: `data.value` is unknown

tonyhallett: I know that. For this particular piece of code why have you cast to number ? What is the type of the value of data.value ?

src/container/container.ts
```diff
@@ -408,7 +408,7 @@ class Container implements interfaces.Container {
     if (Reflect.hasMetadata(METADATA_KEY.PRE_DESTROY, constructor)) {
       const data: interfaces.Metadata = Reflect.getMetadata(METADATA_KEY.PRE_DESTROY, constructor);
 
-      return instance[data.value]();
+      return instance[data.value as number]();
```
tonyhallett: again why the cast to number ?

src/planning/target.ts
```diff
@@ -1,25 +1,25 @@
...
-  public metadata: Metadata<T>[];
+  public metadata!: Metadata[];
 
   public constructor(
     type: interfaces.TargetType,
     identifier: string | symbol,
-    serviceIdentifier: interfaces.ServiceIdentifier<T>,
-    namedOrTagged?: (string | Metadata<T>)
+    serviceIdentifier: interfaces.ServiceIdentifier,
+    namedOrTagged?: (string | unknown)
```
tonyhallett: `string | Metadata`

test/bugs/issue_633.test.ts
```diff
@@ -22,7 +22,7 @@ describe("Issue 633", () => {
     container.bind<Logger>(TYPE.Logger).toDynamicValue((context) => {
       const namedMetadata = context.currentRequest.target.getNamedTag();
```
tonyhallett, line 23: Perhaps we could make interface Metadata generic - `Metadata<TValue>` such that getNamedTag returns `Metadata<string>`

src/container/container.ts
PodaruDragos: Type 'unknown' cannot be used as an index type.ts(2538)

src/planning/reflection_utils.ts
```diff
@@ -196,7 +203,7 @@ function getBaseClassDependencyCount(metadataReader: interfaces.MetadataReader,
 
     // get unmanaged metadata
     const metadata: any[] = targets.map((t: interfaces.Target) =>
-      t.metadata.filter((m) =>
+      t.metadata.filter((m: interfaces.Metadata) =>
```
tonyhallett: you were ok to remove the typing

PodaruDragos:
I pushed another commit where I replaced most of the `any's`. 
Can someone else please also work on this ? 
I don't think I have more time to dedicate to finish this.

@notaphplover  @tonyhallett ?

PodaruDragos:
> > Type 'unknown' cannot be used as an index type.ts(2538)
> 
> I know that ! Why are you casting to number ?!

Oh sorry misunderstood your point. You are right it should be a `string`

PodaruDragos:
I think this PR looks reasonable to by how much amount of work has been done to it. Please let's try to work together on this so we can get closer to moving to `eslint` and be done with this kinds of improvements and really focus on what matters ( and that is v6 BindingAbstraction and other features like so )

tonyhallett:
I don't have the time either at the moment.
I will change what I think needs to change from my comments above.

PodaruDragos:
also @tonyhallett thank you for taking the time with me on this. it's been a long journey on this PR

PodaruDragos:
> I will change what I think needs to change from my comments above.

thanks

tonyhallett:
@PodaruDragos 

Please can you correct what I have identified in https://github.com/inversify/InversifyJS/commit/562243c343423249c9e83263b8b60c4226b54a7d.

I would suggest that no further changes are attempted on this pull request.

Additional changes / unanswered review comments

[Make Metadata generic](https://github.com/inversify/InversifyJS/pull/1311#discussion_r700991244)
[Unnecessary? Typing the decorator](https://github.com/inversify/InversifyJS/pull/1311#discussion_r697340724)


[different tsconfigs for node / browser](https://github.com/inversify/InversifyJS/pull/1311#issuecomment-907092941)

tonyhallett:
@PodaruDragos 
Please can you tell me what comment of mine you deleted

PodaruDragos:
> @PodaruDragos
> Please can you tell me what comment of mine you deleted

yeah sorry , the one with cast to number that I modified to be as string.

tonyhallett:
> yeah sorry , the one with cast to number that I modified to be as string.

Ok :wink:

wiki/provider_injection.md
```diff
@@ -58,7 +58,7 @@ The signature of a provider look as follows:
 
 ```ts
 interface Provider<T> extends NewableFunction {
-    (...args: any[]): (((...args: any[]) => Promise<T>) | Promise<T>);
+    (...args: unknown[]): (((...args: unknown[]) => Promise<T>) | Promise<T>);
```
tonyhallett, line 61: Should we replace the interface with the type that we are exporting ? Similarly with ProviderCreator.

src/interfaces/interfaces.ts
```diff
@@ -71,10 +71,18 @@ namespace interfaces {
     cache: null | TActivated | Promise<TActivated>;
   }
 
-  export type Factory<T, U extends unknown[] = unknown[], V extends unknown[] = unknown[]> = (...args: U) => (((...args: V) => T) | T);
+  export type SimpleFactory<T, U extends unknown[] = unknown[]> =  (...args: U) => T;
+
+  export type MultiFactory<T, U extends unknown[] = unknown[], V extends unknown[] = unknown[]> = (...args: U) => SimpleFactory<T,V>
```
tonyhallett: @PodaruDragos 
Helper types added.  We need a better name for MultiFactory

tonyhallett: We need some of the markdown files to be generated from actual tests so that required updates to markdown automatically occur when we change types.  Will also prevent incorrect typing that I encountered when adding new functionality.

PodaruDragos: what about `CompoundFactory` ? 
I also think `MultiFactory` is fine

PodaruDragos:
> Please revert / adjust accordingly
> `getDependencies`
> `getBaseClassDependencyCount`
> `listRegisteredBindingsForServiceIdentifier`

this should also be done

tonyhallett:
> Please revert / adjust accordingly
getDependencies
getBaseClassDependencyCount
listRegisteredBindingsForServiceIdentifier

> this should also be done

You did two of them, I did the other.

PodaruDragos:
> cntnr should be cast to object with _bindingDictionary

why are we not making this public ?

PodaruDragos:
> You did two of them, I did the other.

oh sorry, I must have slipped that

tonyhallett:
> why are we not making this public ?

We want to control access through fluent syntax

PodaruDragos:
also we have a lot of tests where we do this

`// cast to any to be able to access private props`. maybe we could consider making those public aswell, since we are not really using `private methods`. or change  everything to be `private methods` ex: `#example = () =>  void`

tonyhallett:
> also we have a lot of tests where we do this

> // cast to any to be able to access private props. maybe we could consider making those public aswell, since we are not really using private methods. or change everything to be private methods ex: #example = () => void

The binding tests are poor.
There are multiple tests of this form
```
  it("Should set its own properties correctly", () => {

    interface Ninja { }
    const ninjaIdentifier = "Ninja";

    const binding = new Binding<Ninja>(ninjaIdentifier, BindingScopeEnum.Transient);
    
    const bindingInSyntax = new BindingInSyntax<Ninja>(binding);

    // cast to any to be able to access private props
    const _bindingInSyntax: any = bindingInSyntax;

    expect(_bindingInSyntax._binding.serviceIdentifier).eql(ninjaIdentifier);

  });
```
Why are we checking the serviceIdentifier ! Should be === on the binding.
Also for BindingWhenOnSyntax and BindingInWhenOnSyntax the _binding field is not used !
The test for BindingInSyntax is not required as we are testing that the scope gets changed !

binding_when_on_syntax.test.ts and binding_in_when_on_syntax.test.ts

Both of these just pass on, yet they do not test that they call with the correct parameter or return correctly !

---
Given that these types are never exposed there is no reason why we cannot make them public readonly

Also given that these are all exposed with interfaces is there any reason why BindingToSyntax isn't the only binding syntax class ?

PodaruDragos:
> Also given that these are all exposed with interfaces is there any reason why BindingToSyntax isn't the only binding syntax class ?

I don't really understand what you are asking here.

I have one more thing I want to propose. What if we do this ` export type Newable<T> = new (...args: never) => T;` this way we drop the `any` and it behaves exactly the same ( we still don't have any typesafety for the `...args` ) and then where we need to or overlap with other `...args` we just cast them from `unknown` to `never`. I think just `toConstructor` does that. What are your thoughts on this ?

Also do you think there is a way to properly cast `target` in `injectable` , `decorate` , `_decorate` just so it's not `any` anymore.


Also regarding `@multipleMetadataDecorator` 

```
    @inject<Thing>("Thing")
    @inject("Thing") thing: Thing
```

Is this not more or less the same thing ?

tonyhallett:
> Also given that these are all exposed with interfaces is there any reason why BindingToSyntax isn't the only binding syntax class ?

I don't really understand what you are asking here.

I am saying why have separate binding syntax classes.  Why not put all functionality in a single class that inherits all interfaces - BindingToSyntax given that it is the entry.  The return values are interfaces and ensure correct call order.  The single class just returns itself for the next call.

> Also regarding @multipleMetadataDecorator

```
    @inject<Thing>("Thing")
    @inject("Thing") thing: Thing

```

I assume that you mean replace injecting in setter with injecting the field.
I am injecting in a setter for a reason.  So they are not the same thing at all.

> export type Newable<T> = new (...args: never) => T; 

Why are you doing this ?
Have you tried it throughout the code ?

toConstructor should really be
`toConstructor(constructor: T)`
but we do ` toConstructor<T2>(constructor: Newable<T2>):` so that it can be newed.

I think that we should do this
```
public toConstructor<T2>(constructor: interfaces.Newable<T2>)): interfaces.BindingWhenOnSyntax<T> {
    this._binding.type = BindingTypeEnum.Constructor;
    this._binding.implementationType = constructor as unknown as T;
```

Going to have to come back to you 

> Also do you think there is a way to properly cast target in injectable , decorate , _decorate just so it's not any anymore.

PodaruDragos:
> I assume that you mean replace injecting in setter with injecting the field.
> I am injecting in a setter for a reason. So they are not the same thing at all.

my apologies, didn't actually check the code ( I was curios just about the typings )

> I am saying why have separate binding syntax classes. Why not put all functionality in a single class that inherits all interfaces - BindingToSyntax given that it is the entry. The return values are interfaces and ensure correct call order. The single class just returns itself for the next call.

That would be better in every way, yes.



> Why are you doing this ?
> Have you tried it throughout the code ?

Yes seems to work, will have to double check tough.

> Going to have to come back to you

looking forward

PodaruDragos:
@tonyhallett I have pushed another commit ( I honestly don't know how to get rid of the remaining `any's`.
Do you have time to maybe look at that ?

I also resolved your feedback.
Let me know what you think.

tonyhallett:
@PodaruDragos 

Going to be busy for a few days.  Will have a look soon.

jkhmtx:
@PodaruDragos In case it adds any context, this looks to me like where the error in the test case may be originating from:
image

It looks like `_createSubRequests` is failing as expected given the nature of the test case, but that `context.plan` is `undefined` for some reason.

Looking at the body of `_createSubRequests`, it appears that what could be throwing is the invocation of `_getActiveBindings`. 

image

I'm inferring this because we use `context.addPlan` after that invocation, which would prevent `context.plan` from being `undefined` in the parent scope. I'm a little in the weeds trying to understanding the `_getActiveBindings` method, so I'm not sure how much more investigation I can manage.

An interesting final piece of context, it seems about five months ago this change was made in the commit history which makes accessing `context.plan.rootRequest` more unsafe:

image

Here is the associated PR: https://github.com/inversify/InversifyJS/pull/1132

I am willing to bet the original code was meant to cover up an intermittent regression in accessing this property, and it was deemed unnecessary in the new PR and was thus removed. Maybe it was not time to move it yet and the root problem is still unsolved, or has returned?

Here is a comment in the PR which discusses the change: https://github.com/inversify/InversifyJS/pull/1132/files#r623122372

cc: @tonyhallett

tonyhallett:
Given that Node 10 passed for target es5 but not es6 and node 12 passes for es6 I was thinking perhaps https://node.green/#ES2015-built-in-extensions-Array-prototype-methods-Array-prototype-values

but the ci was passing before The ci passed before https://github.com/inversify/InversifyJS/pull/1311/commits/e37f8457be9ab727fd39dbdac4c48d25f9c23cde
and that only introduced typing changes.

tonyhallett:
I will look to see if the removal of the code was a bad idea.  It seems like it was.

PR #1331, opened 2021-05-03
Fix constant value deactivation

### LINKED ISSUE #1330: Should not deactivate a constant value that has not been activated
https://github.com/inversify/InversifyJS/blob/9f6f43e53752679a610684baceea98533ad65c4b/src/container/container.ts#L640

Change to 
```
private _deactivateIfSingleton(binding: Binding<any>): Promise<void> | void {
        if (!binding.activated) {
            return;
        }
```

tonyhallett:
fix #1330


test/resolution/resolver.test.ts
```diff
@@ -1358,6 +1358,15 @@ describe("Resolve", () => {
       expect(subject1 === subject2).eql(true);
   });
 
+  it("should not deactivate a non activated constant value", () => {
+    const container = new Container();
+    let deactivated = false;
+    container.bind<string>("ConstantValue").toConstantValue("Constant").onDeactivation(_ => {deactivated = true;})
+    container.unbind("ConstantValue");
+    expect(deactivated).to.equal(false);
+  });
```
notaphplover: I'd suggest using sinnon spies next time, but we can afford not to do it now

notaphplover (approved):
Nice! LGTM

src/container/container.ts
```diff
@@ -637,7 +637,7 @@ class Container implements interfaces.Container {
     }
 
     private _deactivateIfSingleton(binding: Binding<any>): Promise<void> | void {
```
notaphplover, line 639: Maybe this method should be called `_deactivateIfActivated` instead

test/resolution/resolver.test.ts
tonyhallett: but then would have had got access to the binding.

src/container/container.ts
tonyhallett: It is clear that deactivation applies to singletons only with _deactivateIfSingleton.  I think activated is a bad name choice given anything can be "activated" through onActivation.  

The [v6-binding-abstractions](https://github.com/inversify/InversifyJS/pull/1336) change I made hopefully makes it clearer.

```
    private _deactivateIfSingleton(binding: interfaces.Binding<any>): Promise<void> | void {
        const scope = binding.scope;
        if(scope.type === ConfigurableBindingScopeEnum.Singleton){
            const cached = scope.resolved; // discriminated union to access directly
            if (cached === undefined) {
                return;
            }

            if (isPromise(cached)) {
                return cached.then((resolved: any) => this._deactivate(binding, resolved));
            }

            return this._deactivate(binding, cached);
        }

    }

```

test/resolution/resolver.test.ts
notaphplover: Sorry I don't understand your point :( (my english is far to be perfect)

tonyhallett: > Sorry I don't understand your point :( (my english is far to be perfect)

There was no point to my point !

You are correct, I should have done
```

  it("should not deactivate a non activated constant value", () => {
    const deactivationHandler = sinon.spy();
    container.bind<string>("ConstantValue").toConstantValue("Constant").onDeactivation(deactivationHandler)
    container.unbind("ConstantValue");
    // tslint:disable-next-line: no-unused-expression
    expect(deactivationHandler.notCalled).to.be.true;
  });
```

Will change

tonyhallett: Done

notaphplover (approved):

PR #1332, opened 2021-05-03
Type improvements

### LINKED ISSUE #1325: Unnecessary generics
https://github.com/inversify/InversifyJS/blob/094bcd6e0b1a56ed85ac9b4f66410910075ef4fc/src/interfaces/interfaces.ts#L217

```
    export type Unbind = <T>(serviceIdentifier: ServiceIdentifier<T>) => void;

    export type UnbindAsync = <T>(serviceIdentifier: ServiceIdentifier<T>) => Promise<void>;

    export type IsBound = <T>(serviceIdentifier: ServiceIdentifier<T>) => boolean;

```
Note that all of below ( shown changed ) should really be an alias for the Container method
```

    export type Bind = Container["bind"];//<T>(serviceIdentifier: ServiceIdentifier<T>) => BindingToSyntax<T>;

    export type Rebind = Container["rebind"];//<T>(serviceIdentifier: ServiceIdentifier<T>) => BindingToSyntax<T>;

    export type Unbind = Container["unbind"];//<T>(serviceIdentifier: ServiceIdentifier<T>) => void;

    export type UnbindAsync = Container["unbindAsync"];//<T>(serviceIdentifier: ServiceIdentifier<T>) => Promise<void>;

    export type IsBound = Container["isBound"];//<T>(serviceIdentifier: ServiceIdentifier<T>) => boolean;
```

tonyhallett:
fix #1325


notaphplover:
I wonder if we want to type this way, I'd prefer not to use string literals in favor of extracting types onto aliases and referencing those aliases, but it's just my opinnion. Maybe we should do a little research. No argument comes to my mind unless the following one: in general people avoid this type of type references (for example mongodb types use it only when there is no other way to do it), but it's not a real argument

tonyhallett:
The argument for not changing is that these functions are not the Container "bind" methods at all.  They may call through but these functions could  in future have additional arguments.

PR #1335, opened 2021-05-03
Deactivation only for singleton

### LINKED ISSUE #1334: Deactivation errors should apply to request scope too
https://github.com/inversify/InversifyJS/blob/9f6f43e53752679a610684baceea98533ad65c4b/src/resolution/instantiation.ts#L119

Change to 
```
function _validateInstanceResolution(binding: interfaces.Binding<unknown>, constr: interfaces.Newable<unknown>): void {
   const scopeMessage =  `Class cannot be instantiated in ${binding.scope} scope.  Deactivation is for singletons.`
   if (binding.scope !== "Singleton") {
        if (typeof binding.onDeactivation === "function") {
            throw new Error(ON_DEACTIVATION_ERROR(constr.name, scopeMessage));
        }

        if (Reflect.hasMetadata(METADATA_KEY.PRE_DESTROY, constr)) {
            throw new Error(PRE_DESTROY_ERROR(constr.name, scopeMessage));
        }
    }
}

```

tonyhallett:
fix #1334 

I did not suffix the error with 'Deactivation is for singletons.'  Add the suffix ?


src/resolution/instantiation.ts
```diff
@@ -116,13 +116,15 @@ function _postConstruct<T>(constr: interfaces.Newable<T>, instance: T): void | P
...
         if (typeof binding.onDeactivation === "function") {
-            throw new Error(ON_DEACTIVATION_ERROR(constr.name, "Class cannot be instantiated in transient scope."));
+            throw new Error(ON_DEACTIVATION_ERROR(constr.name, scopeErrorMessage));
         }
 
         if (Reflect.hasMetadata(METADATA_KEY.PRE_DESTROY, constr)) {
-            throw new Error(PRE_DESTROY_ERROR(constr.name, "Class cannot be instantiated in transient scope."));
+            throw new Error(PRE_DESTROY_ERROR(constr.name, scopeErrorMessage));
         }
     }
```
notaphplover: We should use `BindingTypeEnum` instead

notaphplover (changes requested):
Just a minor change is requested

src/resolution/instantiation.ts
tonyhallett: Ok

test/resolution/resolver.test.ts
```diff
@@ -1475,6 +1475,37 @@ describe("Resolve", () => {
           .throw("onDeactivation() error in class Destroyable: Class cannot be instantiated in transient scope.");
   });
 
+  it("Should not allow request construction with preDestroy", async () => {
+      @injectable()
+      class Destroyable {
+          @preDestroy()
+          public myPreDestroyMethod() {
+              //
```
notaphplover: if you want to remove the empty body warning, you can return `undefined` instead

notaphplover (approved):
LGTM!

PR #1339, opened 2021-05-07
Fix container resolve wrong container

### LINKED ISSUE #1338: Container.resolve should resolve in that container
https://github.com/inversify/InversifyJS/blob/master/wiki/container_api.md#containerresolvetconstructor-interfacesnewablet-t

> Resolve is like container.get<T>(serviceIdentifier: ServiceIdentifier<T>) but it allows users to create an instance even if no bindings have been declared:

It is like it but is resolved in the Context of a child container.  This is not mentioned in the documentation and I would expect that most people would expect that this is just sugar for below

```
this.bind<T>(constructorFunction).toSelf();
const resolved =  this.get<T>(constructorFunction);
```

In most scenarios the difference is not important but it is if you are expecting context.container to be the container that you called resolved upon.

Soln - resolve from the same container !

tonyhallett:
Fix #1338


notaphplover:
Hi @tonyhallett can I tag this PR as `blocked` until the discussion in the related issue comes to an end?

notaphplover (commented):
LGTM if we take this approach, lets discuss this in the relate issue :), docs and changelog should be updated before merging

notaphplover (commented):
Nice! We only need to add changes to the changelog and we are ready to merge

Issue #944, opened 2018-08-28
LazyServiceIdentifer does not work for property injection


MeirionHughes:
Unable to lazily inject circular service dependency via _property_ injection. 

Side-note: Identifer -> Identifier

## Expected Behavior
Inject without error

## Current Behavior
throws error

## Steps to Reproduce (for bugs)

```ts
import "reflect-metadata";
import { Container, injectable, inject, LazyServiceIdentifer } from "inversify";

const container = new Container();

interface IModuleA {}
interface IModuleB {}

const moduleA = Symbol.for("IModuleA");
const moduleB = Symbol.for("IModuleB");

@injectable()
class ModuleA implements IModuleA {
  @inject(new LazyServiceIdentifer(()=>moduleB))
  b: IModuleB;
}

@injectable()
class ModuleB implements IModuleB{
  @inject(new LazyServiceIdentifer(()=>moduleA))
  a: IModuleA
}

container.bind(moduleA).to(ModuleA).inSingletonScope();
container.bind(moduleB).to(ModuleB).inSingletonScope();

let bar = container.get(moduleA);

console.log(bar);
```

## Context

I'm aware that injecting via the constructor works; however in NInject I was able to do this kind of lazy property injection within a base "Module" class. Subsequent derivations of that "Module" class then didn't have to deal with adding those dependencies manually. 

## Environment
inversify: "^4.13.0"
node-js: 10.1.0

# Stack trace
```
Error: No matching bindings found for serviceIdentifier: [object Object]
    at _validateActiveBindingCount (D:\Code\logger\node_modules\inversify\lib\planning\planner.js:62:23)
    at _getActiveBindings (D:\Code\logger\node_modules\inversify\lib\planning\planner.js:48:5)
    at _createSubRequests (D:\Code\logger\node_modules\inversify\lib\planning\planner.js:91:26)
    at D:\Code\logger\node_modules\inversify\lib\planning\planner.js:115:17
    at Array.forEach (<anonymous>)
    at D:\Code\logger\node_modules\inversify\lib\planning\planner.js:114:26
    at Array.forEach (<anonymous>)
    at _createSubRequests (D:\Code\logger\node_modules\inversify\lib\planning\planner.js:94:20)
    at Object.plan (D:\Code\logger\node_modules\inversify\lib\planning\planner.js:136:9)
    at D:\Code\logger\node_modules\inversify\lib\container\container.js:318:37
```

Halynsky:
[...]

Issue #1342, opened 2021-05-08
only inject decorator can be applied to setters 


tonyhallett:
## Expected Behavior
They should all be applicable to setters

## Current Behavior
Only applicable to a property not accessors.

## Possible Solution
Pull request on its way

tonyhallett:
closed by #1343

PR #1177, opened 2019-12-14
fix error when using LazyServiceIdentifer propetrty injection

csr632:
## Description
LazyServiceIdentifer where added at https://github.com/inversify/InversifyJS/pull/774 . But the PR only supported LazyServiceIdentifer in contruction injection and forgot to consider property injection. 

## Related Issue

fix https://github.com/inversify/InversifyJS/issues/944
fix https://github.com/inversify/InversifyJS/issues/865

## Types of changes

- [x] Updated docs / Refactor code / Added a tests case (non-breaking change)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] I have updated the changelog.

PR #1343, opened 2021-05-08
Improve annotations

tonyhallett:
## Description

I have created a new function that is used for all the decorators that are applicable to parameters and properties ( including accessors ).

```
//e.g
function multiInject(serviceIdentifier: interfaces.ServiceIdentifier<any>) {
  return createTaggedDecorator(new Metadata(METADATA_KEY.MULTI_INJECT_TAG, serviceIdentifier));
}

function createTaggedDecorator(
    metadata:interfaces.MetadataOrMetadataArray,
    callback?:(target: any, targetKey: string, index?: number | PropertyDescriptor) => void
) {
    return function(target: any, targetKey: string, index?: number | PropertyDescriptor) {
        if(callback){
            callback(target, targetKey, index);
        }
        if (typeof index === "number") {
            tagParameter(target, targetKey, index, metadata);
        } else {
            tagProperty(target, targetKey, metadata);
        }
    };
}

```
I have also exported createTaggedDecorator from inversify.ts to be available for anyone to use.
Note that this function allows multiple metadata to be provided through a single decorator.

I have also refactored _tagParameterOrProperty as it was checking if it was being applied for a parameter or a property.  Now it is passed only the information that it requires.

I also changed the type to include undefined
`let paramOrPropertyMetadata: interfaces.Metadata[] | undefined = paramsOrPropertiesMetadata[key];`
and then checked against undefined.

## Related Issue
#1342 

## Motivation and Context

#1342

## How Has This Been Tested?

Properly

## Checklist:


src/inversify.ts
```diff
@@ -3,6 +3,7 @@ export const METADATA_KEY = keys;
 export { Container } from "./container/container";
 export { BindingScopeEnum, BindingTypeEnum, TargetTypeEnum } from "./constants/literal_types";
 export { AsyncContainerModule, ContainerModule } from "./container/container_module";
+export { createTaggedDecorator } from "./annotation/decorator_utils"
```
tonyhallett, line 6: This works with interfaces.Metadata but as exposing perhaps should also expose the Metadata class due to its toString override ?

src/annotation/inject.ts
```diff
@@ -18,20 +18,11 @@ export class LazyServiceIdentifer<T = any> {
 }
 
 function inject(serviceIdentifier: ServiceIdentifierOrFunc) {
-  return function(target: any, targetKey: string, index?: number | PropertyDescriptor): void {
+  return createTaggedDecorator(new Metadata(METADATA_KEY.INJECT_TAG, serviceIdentifier),target => {
     if (serviceIdentifier === undefined) {
       throw new Error(UNDEFINED_INJECT_ANNOTATION(target.name));
```
tonyhallett: note that there is no such existing error for multi inject

tonyhallett (commented):
Perhaps could permit inject and multi inject to have two addditional parameters as a shortcut for the named and tagged decorators.

src/utils/js.ts
```diff
@@ -0,0 +1,11 @@
+export function getArrayDuplicate<T>(array:T[]):T | undefined {
+  const seenValues: any= {}
+
+  for (const entry of array) {
+    if (seenValues[entry]) {
+      return entry;
+    } else {
+      seenValues[entry] = true
+    }
+  }
```
notaphplover: A `Set` is probably more appropiate for this

src/utils/js.ts
```diff
@@ -0,0 +1,11 @@
...
+}
```
notaphplover: Line feed

src/inversify.ts
notaphplover: It's hard to say, we could expose it but I wonder if that will cause trouble in future refactors

notaphplover (commented):
It looks great, I just added some comments

src/utils/js.ts
tonyhallett: Thanks

src/utils/js.ts
tonyhallett: Ok

notaphplover (commented):
Great, we only need to update the changelog and this PR should be ready :)

src/annotation/decorator_utils.ts
```diff
@@ -1,68 +1,97 @@
 import * as ERROR_MSGS from "../constants/error_msgs";
 import * as METADATA_KEY from "../constants/metadata_keys";
 import { interfaces } from "../interfaces/interfaces";
+import { getFirstArrayDuplicate } from "../utils/js";
 
 function tagParameter(
     annotationTarget: any,
     propertyName: string,
```
notaphplover: this is now `string | undefined` (we want to set strict mode soon)

src/annotation/decorator_utils.ts
```diff
@@ -1,68 +1,97 @@
...
     paramsOrPropertiesMetadata[key] = paramOrPropertyMetadata;
     Reflect.defineMetadata(metadataKey, paramsOrPropertiesMetadata, annotationTarget);
 
 }
 
+function createTaggedDecoratorInternal(
+    metadata:interfaces.MetadataOrMetadataArray,
+    callback?:(target: any, targetKey: string, indexOrPropertyDescriptor?: number | PropertyDescriptor) => void
+) {
+    return function(target: any, targetKey: string, indexOrPropertyDescriptor?: number | PropertyDescriptor) {
```
notaphplover: Looking at `Reflect` types we should allow targetKey to be also a symbol. Should we handle this on the PR as part of the improvement? It would be nice

notaphplover (commented):
Great! I just added some comments :)

src/annotation/decorator_utils.ts
tonyhallett: No. createTaggedDecorator is a decorator factory.

from https://www.typescriptlang.org/docs/handbook/decorators.html

The expression for the accessor decorator will be called as a function at runtime, with the following three arguments:

Either the constructor function of the class for a static member, or the prototype of the class for an instance member.
The name of the member.
The Property Descriptor for the member.

__
The expression for the property decorator will be called as a function at runtime, with the following two arguments:

Either the constructor function of the class for a static member, or the prototype of the class for an instance member.
The name of the member.
__

The expression for the parameter decorator will be called as a function at runtime, with the following three arguments:

Either the constructor function of the class for a static member, or the prototype of the class for an instance member.
The name of the member.
The ordinal index of the parameter in the function’s parameter list.

tonyhallett: Take that back
declare type ClassDecorator = <TFunction extends Function>(target: TFunction) => TFunction | void;
declare type PropertyDecorator = (target: Object, propertyKey: string | symbol) => void;
declare type MethodDecorator = <T>(target: Object, propertyKey: string | symbol, descriptor: TypedPropertyDescriptor<T>) => TypedPropertyDescriptor<T> | void;
declare type ParameterDecorator = (target: Object, propertyKey: string | symbol, parameterIndex: number) => void;

src/annotation/decorator_utils.ts
tonyhallett: Have gone for `string | symbol | undefined` to match the typescript decorator definition.

> declare type ParameterDecorator = (target: Object, propertyKey: string | symbol, parameterIndex: number) => void;

Although when would it be a symbol ?

tonyhallett: Also had to cast to unknown first when calling as cannot add undefined to targetKey

https://github.com/microsoft/TypeScript/issues/33260

tonyhallett:
I have one other commit to come regarding an unhelpful error message that can arise for decorated properties missing inject / multi inject.

src/interfaces/interfaces.ts
```diff
@@ -111,9 +111,7 @@ namespace interfaces {
         setCurrentRequest(request: Request): void;
     }
 
-    export interface ReflectResult {
```
tonyhallett, line 114: This was deleted as targetKey has changed to string | symbol.

Note that there was already
```
    export interface MetadataMap {
        [propertyNameOrArgumentIndex: string]: Metadata[];
    }
```

but !
```
export interface MetadataReader {
        //..
        getPropertiesMetadata(constructorFunc: Function): MetadataMap;
    }
class MetadataReader implements interfaces.MetadataReader {
    public getPropertiesMetadata(constructorFunc: Function): interfaces.MetadataMap {
        // User generated properties annotations
        const userGeneratedMetadata =  Reflect.getMetadata(METADATA_KEY.TAGGED_PROP, constructorFunc) || [];
        return userGeneratedMetadata;
    }

}

//decorator_utils
function _tagParameterOrProperty(
    metadataKey: string,
    annotationTarget: any,
    key: string | symbol,
    metadata: interfaces.MetadataOrMetadataArray,
) {
    const metadatas: interfaces.Metadata[] = _ensureNoMetadataKeyDuplicates(metadata);

    let paramsOrPropertiesMetadata:any = {};
    // read metadata if available
    if (Reflect.hasOwnMetadata(metadataKey, annotationTarget)) {
        paramsOrPropertiesMetadata = Reflect.getMetadata(metadataKey, annotationTarget);
    }

    let paramOrPropertyMetadata: interfaces.Metadata[] | undefined = paramsOrPropertiesMetadata[key];

    if (paramOrPropertyMetadata === undefined) {
        paramOrPropertyMetadata = [];
    } else {
        for (const m of paramOrPropertyMetadata) {
            if (metadatas.some(md => md.key === m.key)) {
                throw new Error(`${ERROR_MSGS.DUPLICATED_METADATA} ${m.key.toString()}`);
            }
        }
    }

    // set metadata
    paramOrPropertyMetadata.push(...metadatas);
    //********** paramOrPropertiesMetadata is not a MetadataMap !!!
    paramsOrPropertiesMetadata[key] = paramOrPropertyMetadata;
    Reflect.defineMetadata(metadataKey, paramsOrPropertiesMetadata, annotationTarget);

}
```

Something is amiss !

tonyhallett: The result of above is that in reflection_utils we do not get a Target for symbol properties with inject / multiInject decorator due to Object.keys.
```
function getClassPropsAsTargets(metadataReader: interfaces.MetadataReader, constructorFunc: Function, constructorName:string) {

    const classPropsMetadata = metadataReader.getPropertiesMetadata(constructorFunc);
    let targets: interfaces.Target[] = [];
    const keys = Object.keys(classPropsMetadata);
```

I will add another commit that permits this kind of injection.

tonyhallett: Latest commits
permit injection.... this provided the missing functionality.  If we could change the interface MetadataMap to have key string or symbol then we should ! https://github.com/microsoft/TypeScript/issues/42762 see Generalized index signatures.

move symbol tests to container as browser tests were failing - due to no IE11 support for symbol.

tonyhallett: Target interface and class now have additional identifier property

tonyhallett: I could fix #944 in here too ? There is a pr #1177 for it.

I suspect that multiInject should also accept LazyServiceIdentifier.
```
function multiInject(serviceIdentifier: interfaces.ServiceIdentifier<any>) {
  return createTaggedDecoratorInternal(new Metadata(METADATA_KEY.MULTI_INJECT_TAG, serviceIdentifier));
}
//vs
function inject(serviceIdentifier: ServiceIdentifierOrFunc) {
  return createTaggedDecoratorInternal(new Metadata(METADATA_KEY.INJECT_TAG, serviceIdentifier),target => {
    if (serviceIdentifier === undefined) {
      throw new Error(UNDEFINED_INJECT_ANNOTATION(target.name));
    }
  })
}
```

notaphplover: Hi @tonyhallett I will review the PR in a couple of minutes :), I think we should keep PR as small as possible so I think we should review and progress #1177 instead

src/annotation/decorator_utils.ts
```diff
@@ -1,67 +1,99 @@
...
+    paramOrPropertyMetadata.push(...metadatas);
     paramsOrPropertiesMetadata[key] = paramOrPropertyMetadata;
     Reflect.defineMetadata(metadataKey, paramsOrPropertiesMetadata, annotationTarget);
 
 }
+type ParameterOrPropertyDecorator = (
+    target: any,
+    targetKey: string | symbol,
+    indexOrPropertyDescriptor?: number | TypedPropertyDescriptor<unknown>
+) => void;
```
notaphplover: This type is used implicitly outside this module, should we move it to interfaces module?

src/annotation/decorator_utils.ts
```diff
@@ -1,67 +1,99 @@
...
-    }
-
+    let paramsOrPropertiesMetadata:any = {};
     // read metadata if available
     if (Reflect.hasOwnMetadata(metadataKey, annotationTarget)) {
         paramsOrPropertiesMetadata = Reflect.getMetadata(metadataKey, annotationTarget);
     }
 
-    // get metadata for the decorated parameter by its index
-    let paramOrPropertyMetadata: interfaces.Metadata[] = paramsOrPropertiesMetadata[key];
```
notaphplover: I know Typescript sucks at this point with index signatures, could we use this workaround instead of `any`:

```ts
    let paramsOrPropertiesMetadata: Record<string | symbol, interfaces.Metadata[] | undefined> = {};
    // read metadata if available
    if (Reflect.hasOwnMetadata(metadataKey, annotationTarget)) {
        paramsOrPropertiesMetadata = Reflect.getMetadata(metadataKey, annotationTarget);
    }

    let paramOrPropertyMetadata: interfaces.Metadata[] | undefined = paramsOrPropertiesMetadata[key as string];
```

src/annotation/decorator_utils.ts
```diff
@@ -1,67 +1,99 @@
...
 
 }
+type ParameterOrPropertyDecorator = (
+    target: any,
+    targetKey: string | symbol,
+    indexOrPropertyDescriptor?: number | TypedPropertyDescriptor<unknown>
+) => void;
+function createTaggedDecoratorInternal(
+    metadata:interfaces.MetadataOrMetadataArray,
+    callback?:ParameterOrPropertyDecorator
```
notaphplover: This callback is used only on the `@inject` decorator, we should remove this parameter in favor of creating an arrow in the `@inject` decorator:

```ts
function inject(serviceIdentifier: ServiceIdentifierOrFunc) {
  return (target, targetKey, indexOrPropertyDescriptor) => {
      if (serviceIdentifier === undefined) {
          throw new Error(UNDEFINED_INJECT_ANNOTATION(target.name));
      }

      return createTaggedDecoratorInternal(new Metadata(METADATA_KEY.INJECT_TAG, serviceIdentifier)(target, targetKey, indexOrPropertyDescriptor);
  };
}
```

src/annotation/decorator_utils.ts
```diff
@@ -1,67 +1,99 @@
...
+) {
+    const decorator:ParameterOrPropertyDecorator = (target, targetKey, indexOrPropertyDescriptor) => {
+        if(callback){
+            callback(target, targetKey, indexOrPropertyDescriptor);
+        }
+        if (typeof indexOrPropertyDescriptor === "number") {
+            tagParameter(target, targetKey as unknown as string | symbol | undefined, indexOrPropertyDescriptor, metadata);
```
notaphplover: I think this cast is not necessary

src/annotation/decorator_utils.ts
```diff
@@ -1,67 +1,99 @@
...
     // set metadata
-    paramOrPropertyMetadata.push(metadata);
+    paramOrPropertyMetadata.push(...metadatas);
     paramsOrPropertiesMetadata[key] = paramOrPropertyMetadata;
     Reflect.defineMetadata(metadataKey, paramsOrPropertiesMetadata, annotationTarget);
 
 }
+type ParameterOrPropertyDecorator = (
+    target: any,
+    targetKey: string | symbol,
```
notaphplover: `targetKey` could also be `undefined`

notaphplover (changes requested):
I have an iteration plan on my mind, I prefer to split in little steps (it would be too complex to do everything in one revision), I bring you some changes we can apply :). Let's see if what I have in mind makes sense after this

src/annotation/decorator_utils.ts
tonyhallett: Agreed

src/annotation/decorator_utils.ts
tonyhallett: Agreed

src/annotation/decorator_utils.ts
tonyhallett: Now there is no callback there is no need for it.

src/annotation/decorator_utils.ts
tonyhallett: I don't think that it can to be acceptable as a decorator.

src/annotation/decorator_utils.ts
tonyhallett: It is as target key cannot be undefined for decorator

tonyhallett:
> I have an iteration plan on my mind, I prefer to split in little steps (it would be too complex to do everything in one revision), I bring you some changes we can apply :). Let's see if what I have in mind makes sense after this

No problem. There are 3 changes here. Annotation, the missing error in planning and the omission of symbolic properties from planning and resolution.

src/annotation/decorator_utils.ts
tonyhallett: I think that multiinject should behave the same and throw for missing serviceidentifierorfunc and that it should go in this pr.

src/annotation/decorator_utils.ts
tonyhallett: Interface MetadataMap can change to type alias to the Record

src/annotation/decorator_utils.ts
tonyhallett: Apparently it can, so sure I had ts error yesterday.

src/annotation/decorator_utils.ts
tonyhallett: done

src/annotation/decorator_utils.ts
tonyhallett: `paramsOrPropertiesMetadata[key as any];`
I cannot change MetadataMap due to the MetadataReader interface.

src/annotation/decorator_utils.ts
notaphplover: Sure :)

src/annotation/decorator_utils.ts
notaphplover: `paramsOrPropertiesMetadata[key as any]` would do the trick. It's dirty, I know, we need to wait Typescript developers to fix this. `any` also works but we are trying not to use `any`

src/annotation/inject_base.ts
```diff
@@ -0,0 +1,17 @@
+import { UNDEFINED_INJECT_ANNOTATION } from "../constants/error_msgs";
+import { Metadata } from "../planning/metadata";
+import { createTaggedDecorator } from "./decorator_utils";
+import { ServiceIdentifierOrFunc } from "./lazy_service_identifier";
+
+export function injectBase(metadataKey:string){
+  return (serviceIdentifier: ServiceIdentifierOrFunc) => {
+    return (target:any, targetKey:string | symbol, indexOrPropertyDescriptor?:number | TypedPropertyDescriptor<unknown>) => {
```
notaphplover: `targetKey` can be `undefined`

src/annotation/inject_base.ts
```diff
@@ -0,0 +1,17 @@
...
+      if (serviceIdentifier === undefined) {
+        throw new Error(UNDEFINED_INJECT_ANNOTATION(target.name));
+      }
+      return createTaggedDecorator(
+        new Metadata(metadataKey, serviceIdentifier)
+      )(target, targetKey,indexOrPropertyDescriptor);
+    };
+  }
+}
```
notaphplover: Line feed

src/annotation/decorator_utils.ts
```diff
@@ -1,68 +1,87 @@
...
 }
 
+function createTaggedDecorator(
+    metadata:interfaces.MetadataOrMetadataArray,
+) {
+    return (target:any, targetKey:string | symbol | undefined, indexOrPropertyDescriptor?:number | TypedPropertyDescriptor<unknown>) => {
+        if (typeof indexOrPropertyDescriptor === "number") {
+            tagParameter(target, targetKey, indexOrPropertyDescriptor, metadata);
+        } else {
+            tagProperty(target, targetKey as any, metadata);
```
notaphplover:
```suggestion
            tagProperty(target, targetKey as string | symbol, metadata);
```

notaphplover (commented):
Great! I think we only need to add changes to the changelog and we are ready to merge :)

tonyhallett:
Do you want me to remove the new error logic and injection of symbol properties and create new issues / pull requests for them ?

notaphplover:
> Do you want me to remove the new error logic and injection of symbol properties and create new issues / pull requests for them ?

I think we can progress the PR as is. Small PRs are great but I think it's something we can achieve applying software development methodologies, it's something I'd like to use in the future :)

tonyhallett:
What should we do with 
```

    export interface MetadataReader {
        getPropertiesMetadata(constructorFunc: Function): MetadataMap;
    }
```
Given that symbols can be keys ?

Should the wiki have an example with symbol property injection ?

Are you happy with the name of the additional Target property identifier ?

PodaruDragos:
@tonyhallett @notaphplover 
Is it maybe possible to try to change `target` from `any` to `unknown` ?
Would be helpful for when we do migrate to `eslint`

tonyhallett:
This is from lib.es5.d.ts
```
declare type PropertyDecorator = (target: Object, propertyKey: string | symbol) => void;
declare type ParameterDecorator = (target: Object, propertyKey: string | symbol, parameterIndex: number) => void;
```
and the Reflect method that we call from the annotations are also Object.

hasOwnMetadata / getMetadata / defineMetadata

We use target.name though

PodaruDragos:
@tonyhallett I know.. I mean here: 

`
function createTaggedDecorator(
    metadata:interfaces.MetadataOrMetadataArray,
    callback?:(target: any, targetKey: string, index?: number | PropertyDescriptor) => void
) {
    return function(target: any, targetKey: string, index?: number | PropertyDescriptor) 
`

tonyhallett:
> @tonyhallett I know.. I mean here:
> 
> `function createTaggedDecorator( metadata:interfaces.MetadataOrMetadataArray, callback?:(target: any, targetKey: string, index?: number | PropertyDescriptor) => void ) { return function(target: any, targetKey: string, index?: number | PropertyDescriptor) `

I know... and injectBase as well

tonyhallett:
@PodaruDragos should be ok now.  Now no undefined message for a property. Good spot

PodaruDragos:
@tonyhallett  that introduces another problem for `eslint`.
I know we can disable some rules to our fitting but using `Object` will throw this error:

```
Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead
```

[link](https://github.com/typescript-eslint/typescript-eslint/blob/v4.22.0/packages/eslint-plugin/docs/rules/ban-types.md)

src/planning/target.ts
```diff
@@ -10,19 +10,23 @@ class Target implements interfaces.Target {
...
+        identifier: string | symbol,
         serviceIdentifier: interfaces.ServiceIdentifier<any>,
         namedOrTagged?: (string | Metadata)
     ) {
 
         this.id = id();
         this.type = type;
         this.serviceIdentifier = serviceIdentifier;
-        this.name = new QueryableString(name || "");
+        const queryableName = typeof identifier === 'symbol' ? identifier.toString().slice(7,-1) : identifier;
```
notaphplover: why should we change the behavior?

notaphplover (commented):
> What should we do with
> 
> ```
> 
>     export interface MetadataReader {
>         getPropertiesMetadata(constructorFunc: Function): MetadataMap;
>     }
> ```
> 
> Given that symbols can be keys ?

I would not change it. It's something Typescript has to fix in order to allow symbol index signatures

> 
> Should the wiki have an example with symbol property injection ?

Probably, but that's something we could add in another PR

> 
> Are you happy with the name of the additional Target property identifier ?

I'm not sure, do we need it? Maybe I'm missing something

src/planning/target.ts
tonyhallett: We were originally getting the property name for QueryableString, this is not sufficient anymore as it can be a symbol now.
We could pass the symbol to QueryableString but we would then need to change the value method to return `string | symbol`.

Looking at the image below the cast should have been to `Record<string | symbol, unknown>`

image

notaphplover: Understood :), then I would suggest to extract `identifier.toString().slice(7,-1)` to a util function :)

src/annotation/inject_base.ts
```diff
@@ -0,0 +1,19 @@
+import { UNDEFINED_INJECT_ANNOTATION } from "../constants/error_msgs";
+import { Metadata } from "../planning/metadata";
+import { createTaggedDecorator } from "./decorator_utils";
+import { ServiceIdentifierOrFunc } from "./lazy_service_identifier";
+
+export function injectBase(metadataKey:string){
+  return (serviceIdentifier: ServiceIdentifierOrFunc) => {
+    return (target: Object,
```
notaphplover: I think `Record<string | symbol, unknown> | interfaces.Newable<unknown>` is a better type for `target`

tonyhallett: Not acceptable by the compiler

> is not assignable to type 'Record<string | symbol, unknown>'. Index signature is missing in type

I think that I have provided a solution

src/planning/target.ts
tonyhallett: >  cast should have been to Record<string | symbol, unknown>

But you cannot index with symbol so will will remain any

tonyhallett: >  extract identifier.toString().slice(7,-1) to a util function :)

done

ok with the property name ?

src/annotation/inject_base.ts
tonyhallett:
```
interface ConstructorFunction {
    new(...args:any[]):unknown,
    readonly prototype:unknown
}
interface Prototype {
    constructor: Function;
}
export type DecoratorTarget = Prototype | ConstructorFunction;
```

notaphplover: It's indeed a valid solution. We want to introduce some linter rules in the future and that's the reason to [ban some types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md). If you want I can create a PR to your branch with equivalent types which respect this rule :)

src/planning/target.ts
notaphplover: Yeah :)

src/annotation/inject_base.ts
tonyhallett: Can you elaborate on CI ?

By all means change the typing. I hadn't seen the banning of Function.  I look forward to seeing the equivalent types.

tonyhallett: > Can you elaborate on CI ?

I assume you mean I. So go ahead !

notaphplover: Yeah, that was a typo

notaphplover: I spent a couple of hours wondering how to improve typings because my original idea didn't work. At the end `object` is good enough imo, it covers all the cases and respects the `Reflect` typings. Sorry for the waste of time

notaphplover (approved):

src/annotation/inject_base.ts
tonyhallett: But in the linked "ban some types" the object type is problematic too ( which I was purposely avoiding )

Also, although object ( and Object ) is allowed for Reflect typings, what we are adding metadata to is the constructor function and as such the annotationTarget should be Function in `_tagParameterOrProperty`.

```
function _tagParameterOrProperty<T>(
    metadataKey: string,
    annotationTarget: object,
    key: string | symbol,
    metadata: interfaces.MetadataOrMetadataArray,
) {
    //....
    Reflect.defineMetadata(metadataKey, paramsOrPropertiesMetadata, annotationTarget);

}
class MetadataReader implements interfaces.MetadataReader {

    public getConstructorMetadata(constructorFunc: Function): interfaces.ConstructorMetadata {

        // TypeScript compiler generated annotations
        const compilerGeneratedMetadata = Reflect.getMetadata(METADATA_KEY.PARAM_TYPES, constructorFunc);

        // User generated constructor annotations
        const userGeneratedMetadata = Reflect.getMetadata(METADATA_KEY.TAGGED, constructorFunc);

        return {
            compilerGeneratedMetadata,
            userGeneratedMetadata: userGeneratedMetadata || {}
        };

    }

    public getPropertiesMetadata(constructorFunc: Function): interfaces.MetadataMap {
        // User generated properties annotations
        const userGeneratedMetadata =  Reflect.getMetadata(METADATA_KEY.TAGGED_PROP, constructorFunc) || [];
        return userGeneratedMetadata;
    }

}

```
 
We then need to ensure that tagParameter passes a function.  I suggest that we change tagParameter to below.
We do not need to check the property to know that the decorator is on a method parameter as the annotation target tells us.
```
function tagParameter(
    annotationTarget: object,
    parameterIndex: number,
    metadata: interfaces.MetadataOrMetadataArray
) {
    if ( typeof annotationTarget === "function") {
        _tagParameterOrProperty(METADATA_KEY.TAGGED, annotationTarget, parameterIndex.toString(), metadata);
    }else{
        throw new Error(ERROR_MSGS.INVALID_DECORATOR_OPERATION);
    }
}
```

src/annotation/decorator_utils.ts
```diff
@@ -72,22 +98,22 @@ function _param(paramIndex: number, decorator: ParameterDecorator) {
 }
 
 // Allows VanillaJS developers to use decorators:
-// decorate(injectable("Foo", "Bar"), FooBar);
+// decorate(injectable(), FooBar);
 // decorate(targetName("foo", "bar"), FooBar);
 // decorate(named("foo"), FooBar, 0);
 // decorate(tagged("bar"), FooBar, 1);
 function decorate(
```
tonyhallett, line 128: Although for js, shouldn't this be overloaded ?

tonyhallett (commented):
Please see https://github.com/inversify/InversifyJS/pull/1343#discussion_r631199923.

src/annotation/inject_base.ts
tonyhallett: The name is required for static method parameters. Take that back.

PodaruDragos: we can also just add [this ](https://palantir.github.io/tslint/rules/ban-types/)into out `tslint` setup right now and then just try out what works best, since `ban-types` are somehow related to both linters.

What was the problem with `Record<string, unknown` anyway ?

tonyhallett: > What was the problem with Record<string, unknown anyway ?

When applied to a property typescript will complain

> is not assignable to type 'Record<string | symbol, unknown>'. Index signature is missing in type CLASSNAME

This is why
```
class ClassWithDecorators {
  constructor(@decorator param:Dependency){}
  @decorator
  //@ts-ignore
  prop:Dependency

  @decorator
  static staticProp:Dependency

}
// here is the error
const typescriptError:Record<string,unknown> = ClassWithDecorators.prototype

```

the js
```
var ClassWithDecorators = /** @class */ (function () {
    function ClassWithDecorators(param) {
    }
    __decorate([
        decorator,
        __metadata("design:type", Dependency)
    ], ClassWithDecorators.prototype, "prop", void 0);
    __decorate([
        decorator,
        __metadata("design:type", Dependency)
    ], ClassWithDecorators, "staticProp", void 0);
    ClassWithDecorators = __decorate([
        __param(0, decorator),
        __metadata("design:paramtypes", [Dependency])
    ], ClassWithDecorators);
    return ClassWithDecorators;
}());
```

tonyhallett: They differ.  object is allowed in tslint.

https://github.com/typescript-eslint/typescript-eslint/blob/1c1b572c3000d72cfe665b7afbada0ec415e7855/packages/eslint-plugin/src/rules/ban-types.ts#L99

https://github.com/palantir/tslint/blob/285fc1db18d1fd24680d6a2282c6445abf1566ee/src/configs/recommended.ts#L23

tonyhallett: @PodaruDragos 
tell eslint to ignore the object and the Function types where they are used.  The errors is there for lazy typing and errors that arise because of that.  
In this instance the type signatures are ok and definitely better than unknown and of couse any.

tonyhallett: If the class being decorated had
`[key:string]:unknown`
Then `Record<string | symbol, unknown> | interfaces.Newable<unknown>` will not error

notaphplover (approved):

src/annotation/decorator_utils.ts
notaphplover: It would be nice, yeah

src/annotation/decorator_utils.ts
```diff
@@ -3,28 +3,40 @@ import * as METADATA_KEY from "../constants/metadata_keys";
...
+function targetIsConstructorFunction<T = Object>(target:DecoratorTarget<T>): target is ConstructorFunction<T>{
+    return (target as any).prototype !== undefined;
+}
+
+interface ConstructorFunction<T = Object>{
+    new (...args:unknown[]): T,
+    prototype:T
+}
+
+export type DecoratorTarget<T = Object> = ConstructorFunction<T> |  T
```
tonyhallett: I believe that this is the best way of typing the target.
We differentiate between constructor function and prototype to prevent static properties and the prototype property is the proper way of doing so.
`_tagParameterOrProperty` requires `annotationTarget: Function` ( this cannot be typed better with generics ) and when target is a constructor function it is obviously so.  Although it is through a cast in `tagParameter` `_tagParameterOrProperty(METADATA_KEY.TAGGED, annotationTarget as ConstructorFunction, parameterIndex.toString(), metadata);` the type predicate could have been used instead.

src/annotation/decorator_utils.ts
```diff
@@ -1,68 +1,110 @@
...
 import { interfaces } from "../interfaces/interfaces";
+import { getFirstArrayDuplicate } from "../utils/js";
+
+function targetIsConstructorFunction<T = Object>(target:DecoratorTarget<T>): target is ConstructorFunction<T>{
+    return (target as any).prototype !== undefined;
+}
+
+interface ConstructorFunction<T = Object>{
+    new (...args:unknown[]): T,
+    prototype:T
```
notaphplover: I think prototype should not be `T` since T is not an instance of the type, but the prototype instead

src/annotation/decorator_utils.ts
```diff
@@ -1,68 +1,110 @@
 import * as ERROR_MSGS from "../constants/error_msgs";
 import * as METADATA_KEY from "../constants/metadata_keys";
 import { interfaces } from "../interfaces/interfaces";
+import { getFirstArrayDuplicate } from "../utils/js";
+
+function targetIsConstructorFunction<T = Object>(target:DecoratorTarget<T>): target is ConstructorFunction<T>{
+    return (target as any).prototype !== undefined;
```
notaphplover:
```suggestion
    return (target as ConstructorFunction<T>).prototype !== undefined;
```

src/annotation/decorator_utils.ts
```diff
@@ -1,68 +1,110 @@
 import * as ERROR_MSGS from "../constants/error_msgs";
 import * as METADATA_KEY from "../constants/metadata_keys";
 import { interfaces } from "../interfaces/interfaces";
+import { getFirstArrayDuplicate } from "../utils/js";
+
+function targetIsConstructorFunction<T = Object>(target:DecoratorTarget<T>): target is ConstructorFunction<T>{
```
notaphplover, line 6:
```suggestion
function targetIsConstructorFunction<T = Object>(target: unknown): target is ConstructorFunction<T>{
```

notaphplover: At this point, are we sure target is a DecoratorTarget<T>? If that's the case we can avoid this

tonyhallett: Why ?

src/annotation/decorator_utils.ts
tonyhallett: agreed

src/annotation/decorator_utils.ts
tonyhallett: image

tonyhallett:
```
var ClassWithDecorators = /** @class */ (function () {
    function ClassWithDecorators(param) {
    }
    __decorate([
        decorator,
        __metadata("design:type", Dependency)
    ], ClassWithDecorators.prototype, "prop", void 0);
    __decorate([
        decorator,
        __metadata("design:type", Dependency)
    ], ClassWithDecorators, "staticProp", void 0);
    ClassWithDecorators = __decorate([
        __param(0, decorator),
        __metadata("design:paramtypes", [Dependency])
    ], ClassWithDecorators);
    return ClassWithDecorators;
}());
```

src/annotation/decorator_utils.ts
tonyhallett: It is known and it certainly is not undefined or null

src/annotation/decorator_utils.ts
tonyhallett: I will leave this one as is for the moment.

src/annotation/decorator_utils.ts
tonyhallett: > At this point, are we sure target is a DecoratorTarget? If that's the case we can avoid this

Avoid the change to unknown ?

We are sure it is a decorator target as the decorator is called with the decorator target.  tagProperty is called with the decorator target and passes it to the type predicate

src/annotation/decorator_utils.ts
tonyhallett: Mapped T to `Prototype<T>`, properties now optional.

```
type Prototype<T> = {
    [Property in keyof T ]:
        T[Property] extends Function?
            T[Property] :
            T[Property] | undefined
} & {constructor:Function}

interface ConstructorFunction<T = Object>{
    new (...args:unknown[]): T,
    prototype:Prototype<T>
}

export type DecoratorTarget<T = Object> = ConstructorFunction<T> |  Prototype<T>
```

PodaruDragos (approved):
This is definitely better ( nice work). feel free to merge when able. LGTM.

tonyhallett:
@PodaruDragos 
@notaphplover 

Thanks for your contributions on this and other pull requests !

Add this to the changelog as exports the helper createTaggedDecorator ?

Are we doing merge commits or rebase and merge ?

PR #1345, opened 2021-05-13
Allow multiple tags while resolving values

### LINKED ISSUE #1110: Complex contextual resolution (multiple tags/tag and name)
In the system I'm building, we have `ViewControllers` and `ViewControllerComponents`. Each ViewController type has a name, like `form_view`, `map_view`, etc. Each ViewController has specific ViewControllerComponents that correspond to it, like `TextArea` and `Numeric` for `form_view`, and `Point` and `Polygon` for `map_view`.

## Expected Behavior
`ViewControllers` are easy:

```typescript
context.bind(VIEW_TYPES.ViewController).to(FormViewController).whenTargetNamed("form_view")
```

And then for the factory I do:
```typescript
bind<ViewControllerFactory<any>>(VIEW_TYPES.ViewControllerFactory).toFactory<View<any>>(
    context => name => context.container.getNamed(VIEW_TYPES.ViewController, name),
);
```

Now what I want to do is the following for `ViewControllerComponents`:
```typescript
context.bind(VIEW_TYPES.ViewControllerComponent).to(TextArea).whenTargetNamed("text_area").whenTargetTagged("view_type", "form_view")
...
bind<ViewControllerComponentFactory<any>>(VIEW_TYPES.ViewControllerComponentFactory).toFactory<ViewComponent<any>>(
    context => view_type => component_name => context.container.getMultiTagged(VIEW_TYPES.ViewControllerComponent,{[TAGS.NAME_TAG]: component_name, "view_type": view_name})
);
```

There are two problems: the first is that you can't chain `bindingToSyntax` (so `.whenTargetNamed().whenTargetTagged()` won't work). I solved this with the following complex constraint:

```typescript
    const constraint: interfaces.ConstraintFunction =  (request: interfaces.Request | null) =>
        request !== null && request.target !== null && (
            request.target.matchesTag("view_type")(view_type)
        ) && (
            names.map(name => request.target.matchesNamedTag(name)).some(i=>i) // check that any array element is truthy
        );
```

But I wasn't able to figure out how to craft a request that can satisfy this request.

## Current Behavior
I dug deeper here and found that inside of the `plan` function in `planner.ts`, a `Target` is created, but you're only allowed to specify a single key for its `Metadata` (despite `Metadata` being an array).

## Possible Solution
If we were somehow allowed to append to this array of `Metadatas`, then what I'm trying to do would be possible. 

Am I missing something, or approaching this the wrong way? There is still a lot about Inversify that I haven't learned yet.

luiz290788:
## Description

This change extends the `getTagged`, `isBoundTagged`, `getAllTagged`, `getTaggedAsync` and `getAllTaggedAsync` to accept multiple tags while resolving the dependency. The functionality was already implemented but not exposed in those methods.

The solution here was to just accept tuples of `[key: symbol | string | number, value: any]` in those methods and covert them into `Metadata` instances while building the plan.

## Related Issue

Closes #1110 

## Motivation and Context

This gives more power while registering instances and resolving them. This is particularly useful when building factories.

## How Has This Been Tested?

This change is covered in unit testing.

## Types of changes

## Checklist:


src/container/container.ts
```diff
@@ -593,15 +592,15 @@ class Container implements interfaces.Container {
...
-        value?: any,
+        tags?: interfaces.Tag[]
     ): GetArgs<T> {
         const getNotAllArgs: GetArgs<T> = {
             avoidConstraints: false,
             isMultiInject,
             serviceIdentifier,
-            key,
-            value,
+            key: tags?.[0]?.[0],
```
luiz290788, line 601: This is to support the current interface of the `GetArgs`. I've marked these attributes as deprecated to support multiple tags in the request.

luiz290788:
Documentation update will be provided shortly.

src/interfaces/interfaces.ts
```diff
@@ -179,6 +182,10 @@ namespace interfaces {
         skipBaseClassChecks?: boolean;
     }
 
+    export type Tag = [key: string | number | symbol, value: any];
```
PodaruDragos: replace the `any` here with `unknown`

src/planning/planner.ts
```diff
@@ -26,18 +26,14 @@ function _createTarget(
     targetType: interfaces.TargetType,
     serviceIdentifier: interfaces.ServiceIdentifier<any>,
     name: string,
-    key?: string | number | symbol,
```
PodaruDragos, line 29: this will be a breaking change, we might want to think about this a bit

luiz290788: Isn't this just a private method?

src/interfaces/interfaces.ts
luiz290788: Done!

tonyhallett:
@PodaruDragos 
There is a problem with this pr.  Will give details shortly.

src/planning/planner.ts
tonyhallett: It is private.  The plan method isn't but it is internal.  Somehow there isn't an existing test in planner.test.ts or resolver.test.ts that tests for this additional metadata.  Perhaps this pr would be an ideal opportunity.

test/inversify.test.ts
```diff
@@ -2808,4 +2808,89 @@ describe("InversifyJS", () => {
 
     });
 
+    describe("Injection of values with multiple tags", () => {
+        it("Should be able to resolve instance", () => {
```
tonyhallett, line 2830: there should be a test for getAllTagged

test/inversify.test.ts
```diff
@@ -2808,4 +2808,89 @@ describe("InversifyJS", () => {
...
+
+            const container = new Container();
+            const throwableConstraint = taggedConstraint('throwable')
+            const spikesConstraint = taggedConstraint('spikes')
+
+            const allConstraints = (...constraints: interfaces.ConstraintFunction[]) =>
+                (request: interfaces.Request) => constraints.every(constraint => constraint(request))
+
+            container.bind<Weapon>("Weapon").toConstantValue(new Shuriken(3))
+                .when(allConstraints(throwableConstraint(true), spikesConstraint(3)))
```
tonyhallett: I think there should be an additional whenTargetMultiTagged

src/container/container.ts
```diff
@@ -620,9 +619,8 @@ class Container implements interfaces.Container {
                 args.isMultiInject,
                 args.targetType,
                 args.serviceIdentifier,
-                args.key,
-                args.value,
-                args.avoidConstraints
+                args.avoidConstraints,
+                args.tags
```
tonyhallett: This does not place nicely with existing middleware

```

        it.only("does not work with middleware", () => {
            const middleware:interfaces.Middleware = (next => {
                return (nextArgs => {
                    if(nextArgs.key === "tag" && nextArgs.value === 1){
                        nextArgs.value = 2;
                    }
                    return next(nextArgs);
                })
            })

            container.applyMiddleware(middleware);
            container.bind('Sid').toConstantValue("One").whenTargetTagged("tag",1);
            container.bind('Sid').toConstantValue("Two").whenTargetTagged("tag",2);
            expect(container.getTagged<string>("Sid","tag",1)).to.equal("Two");
        })
```

tonyhallett (changes requested):

test/inversify.test.ts
luiz290788: I though about creating that, I'll add it.

src/container/container.ts
luiz290788: Oh, because changing key and value in the middleware does not affect the tags therefore the resolution is not affected. Good catch! I'll try to come around that.

src/planning/planner.ts
luiz290788: I'll add those.

test/inversify.test.ts
luiz290788: Quick question about this.

When I use the `whenTargetMultiTagged` does the depdency should only be resolved when trying to resolve a dependency with all the tags or if it matches a single tag it is already enough to resolve that dependency? Basically my question here is if we should use `and` or `or` for the tags condition.

tonyhallett: It is a good point.
If there was to be just this change then a boolean would work.
Although we need to consider the parent and ancestor versions of these.

We could change below to accept multiple tags in a non breaking manner as you did for getTagged etc
These would be or - when having one of these tags
whenTargetTagged
whenParentTagged
whenAnyAncestorTagged
whenNoAncestorTagged

Then a multi tagged version for each of them - these being and.

@PodaruDragos 
@notaphplover 

Your thoughts gentlemen.

luiz290788: I've implemented the `whenTargetMultiTagged` with the `or` behaviour on the definition but forcing the value to have all the tags present in the request.

src/interfaces/interfaces.ts
```diff
@@ -302,6 +309,7 @@ namespace interfaces {
         whenTargetNamed(name: string | number | symbol): BindingOnSyntax<T>;
         whenTargetIsDefault(): BindingOnSyntax<T>;
         whenTargetTagged(tag: string | number | symbol, value: any): BindingOnSyntax<T>;
+        whenTargetMultiTagged(...tags: [interfaces.Tag, ...interfaces.Tag[]]): BindingOnSyntax<T>;
```
luiz290788:
```Typescript
(...tags: [interfaces.Tag, ...interfaces.Tag[]])
```

This type forces the presence of at least one tag.

src/syntax/binding_when_syntax.ts
```diff
@@ -39,6 +39,15 @@ class BindingWhenSyntax<T> implements interfaces.BindingWhenSyntax<T> {
         return new BindingOnSyntax<T>(this._binding);
     }
 
+    public whenTargetMultiTagged(...tags: [interfaces.Tag, ...interfaces.Tag[]]): interfaces.BindingOnSyntax<T> {
+        this._binding.constraint = (request: interfaces.Request) =>
```
tonyhallett: There is an issue with the current typings where the Binding class and the Binding interface differ.

Binding interface
```
constraint: ConstraintFunction;

export interface ConstraintFunction extends Function {
        metaData?: Metadata;
        (request: Request | null): boolean;
}
```

vs - public constraint: (request: interfaces.Request) => boolean;

Although I do not think we call with null it can be through the interface so you need to check for it

You can see the taggedConstraint helper does

```
const taggedConstraint = (key: string | number | symbol) => (value: any) => {

    const constraint: interfaces.ConstraintFunction =  (request: interfaces.Request | null) =>
        request !== null && request.target !== null && request.target.matchesTag(key)(value);

    constraint.metaData = new Metadata(key, value);

    return constraint;
};
```

tonyhallett: This would be a breaking change but should ConstraintFunction change to have metadata property as a Metadata array ?

tonyhallett (changes requested):

src/syntax/binding_when_syntax.ts
tonyhallett: But we could add a multiMetaData property and look for that in https://github.com/inversify/InversifyJS/blob/4332cb8746327cc70b8fa48016cf33786b11909a/src/utils/serialization.ts#L44

src/interfaces/interfaces.ts
tonyhallett: Be better if you could force at least two
(...tags: [interfaces.Tag,interfaces.Tag, ...interfaces.Tag[]])

src/syntax/binding_in_when_on_syntax.ts
```diff
@@ -45,6 +45,10 @@ class BindingInWhenOnSyntax<T> implements interfaces.BindingInSyntax<T>, interfa
         return this._bindingWhenSyntax.whenTargetTagged(tag, value);
     }
 
+    public whenTargetMultiTagged(...tags: [interfaces.Tag, ...interfaces.Tag[]]): interfaces.BindingOnSyntax<T> {
```
tonyhallett: no test for this

src/syntax/binding_when_syntax.ts
tonyhallett: When you change to `(request: interfaces.Request | null)` there needs to be full coverage

tonyhallett (changes requested):

src/syntax/binding_when_syntax.ts
luiz290788: > This would be a breaking change but should ConstraintFunction change to have metadata property as a Metadata array ?

I see, I would potentially be a breaking change if we change that because we expose the `taggedConstraint` with no type defined. 

Although the `metaData` here is only used for the message logging and we could potentially change that. This is something that we only expect users to write and not read. So making it an `Metadata | Metadata[]` is something that is not as harmful.

luiz290788:
I'll work on the tests over the weekend. Thanks for all the fast feedback provided so far!

src/syntax/binding_when_syntax.ts
tonyhallett: > So making it an Metadata | Metadata[] is something that is not as harmful.

Still a breaking change.

It's not that important but for consistenct error messages for bindings with tagged constraints...

When we do v6 we can change to
```
export interface ConstraintFunction extends Function {
        metaData?: Metadata[];
        (request: Request | null): boolean;
}
```

luiz290788: It makes sense. So for now we can keep the metadata null when there is multiple tags on v5 and provide a better error message on v6.

tonyhallett: I would do multiMetadata now personally but I will not block the pr if it is omitted

luiz290788: And introduce the breaking change in v5? Or do we create a new property into `ConstraintFunction`?

tonyhallett: Add optional multiMetadata property.

luiz290788:
I've addressed most of the comments, I still need to add tests for everything.

package.json
```diff
@@ -13,7 +13,7 @@
     "build:es6": "tsc -p src/tsconfig-es6.json",
     "clean": "rm -r amd es lib",
     "pretest": "tslint --project .",
-    "test": "nyc --require ts-node/register mocha test/**/*.test.ts --reporter spec --retries 3 --require 'node_modules/reflect-metadata/Reflect.js' --exit",
+    "test": "nyc --require ts-node/register mocha test/*.test.ts test/**/*.test.ts --reporter spec --retries 3 --require 'node_modules/reflect-metadata/Reflect.js' --exit",
```
luiz290788, line 16: I've realised that we were not running `test/inversify.test.ts` while running tests. I've fixed that to run all the tests in the root folder as well.

luiz290788:
How do I fix the code climate issue? 🤔 

That's a duplication on our code, it is possible to fix that but I think that is a bigger issue. Those two files are almost an entire duplication from each other. I don't think I should be addressing that as part of this pr.

Is there anyway to mark those issues as "won't fix" in code climate for now? I am happy to fix all that duplication in a later pr.

package.json
tonyhallett: They are run in ci.

tonyhallett: But makes sense to run them locally first

tonyhallett:
> How do I fix the code climate issue? 🤔
> 
> That's a duplication on our code, it is possible to fix that but I think that is a bigger issue. Those two files are almost an entire duplication from each other. I don't think I should be addressing that as part of this pr.
> 
> Is there anyway to mark those issues as "won't fix" in code climate for now? I am happy to fix all that duplication in a later pr.

I cannot mark them. @dcavanagh

src/container/container.ts
```diff
@@ -593,15 +592,15 @@ class Container implements interfaces.Container {
...
     ): GetArgs<T> {
         const getNotAllArgs: GetArgs<T> = {
             avoidConstraints: false,
             isMultiInject,
             serviceIdentifier,
-            key,
-            value,
+            key: tags?.[0]?.[0],
+            value: tags?.[0]?.[1],
+            tags: tags?.slice(1)
```
tonyhallett: This is a problem.

```

        it("Should provide all tags in a middleware", () => {
            const middleware:interfaces.Middleware = (next => {
                return (nextArgs => {
                    if(nextArgs.tags!.length !== 2){
                        throw new Error("Where's the tag !?")
                    }
                    return next(nextArgs);
                })
            })

            container.applyMiddleware(middleware);
            expect(() => container.getTagged<string>("Sid", [["tag1", 1],["tag2",2]])).not.to.throw("Where's the tag !?");
        })
```

src/utils/serialization.ts
```diff
@@ -45,6 +45,10 @@ function listRegisteredBindingsForServiceIdentifier(
                 registeredBindingsList = `${registeredBindingsList} - ${binding.constraint.metaData}`;
             }
 
+            if (binding.constraint.multiMetaData) {
+                registeredBindingsList = `${registeredBindingsList} - ${binding.constraint.multiMetaData.join(', ')}`;
```
tonyhallett, line 49: No test

test/syntax/binding_when_on_syntax.test.ts
```diff
@@ -107,6 +109,7 @@ describe("BindingWhenOnSyntax", () => {
         expect(whenStub.callCount).eql(1);
         expect(whenTargetNamedStub.callCount).eql(1);
         expect(whenTargetTaggedStub.callCount).eql(1);
+        expect(whenTargetMultiTaggedStub.callCount).eql(1);
```
tonyhallett, line 112: Being picky but you have followed all the other poor tests that do not test that the correct arguments are passed through.

test/syntax/binding_in_when_on_syntax.test.ts
```diff
@@ -132,6 +134,7 @@ describe("BindingInWhenOnSyntax", () => {
         expect(whenStub.callCount).eql(1);
         expect(whenTargetNamedStub.callCount).eql(1);
         expect(whenTargetTaggedStub.callCount).eql(1);
+        expect(whenTargetMultiTaggedStub.callCount).eql(1);
```
tonyhallett, line 137: As before. Called with what ?

src/syntax/constraint_helpers.ts
```diff
@@ -27,6 +27,18 @@ const taggedConstraint = (key: string | number | symbol) => (value: any) => {
     return constraint;
 };
 
+const multiTaggedConstraint = (...tags: [interfaces.Tag, ...interfaces.Tag[]]) => {
+    const constraint: interfaces.ConstraintFunction = (request: interfaces.Request) =>
```
tonyhallett, line 31: `Request | null`

src/interfaces/interfaces.ts
```diff
@@ -94,6 +94,7 @@ namespace interfaces {
         serviceIdentifier: interfaces.ServiceIdentifier<T>;
         key?: string | number | symbol;
         value?: any;
+        tags?: Tag[];
```
tonyhallett, line 97: This would be better to not be optional

src/syntax/constraint_helpers.ts
tonyhallett: And full coverage

tonyhallett: There are some uncovered lines in Container.resolve that are mine
https://github.com/inversify/InversifyJS/pull/1347

src/container/container.ts
luiz290788: Do you have any suggestion on how to solve this without breaking the other test you provided?

I've tried the solution with getters and setters but it was a tricky one to solve. I'll invest more time on this.

tonyhallett: May have missed something myself.  First attempt

```

    private _getNotAllArgs<T>(
        serviceIdentifier: interfaces.ServiceIdentifier<T>,
        isMultiInject: boolean,
        tags?: interfaces.Tag[]
    ): GetArgs<T> {
        const getNotAllArgs: GetArgs<T> = {
            avoidConstraints: false,
            isMultiInject,
            serviceIdentifier,
            key: tags?.[0]?.[0],
            value: tags?.[0]?.[1],
            tags: tags??[]
        };

        return getNotAllArgs;
    }



    private _planAndResolve<T>(): (args: interfaces.NextArgs<T>) => interfaces.ContainerResolution<T> {
        return (args: interfaces.NextArgs<T>) => {
            if (args.tags.length> 0 ){
                const tags = [...args.tags];
                const firstTag = tags.shift()
                if(args.key !== firstTag![0] || args.value !== firstTag![1]){
                    if (args.key !== undefined) {
                        args.tags = [[args.key,args.value],...tags]
                    }else {
                        args.tags = tags;
                    }
                }
            }
```

luiz290788: The problem is that if change the array of `tags` and do not change the `key` and `value` this code will override the values with the original value.

tonyhallett: Indeed

tonyhallett: could store the original args on the container for checking in planAndResolve if that helps

tonyhallett: Thinking aloud

```
private _get<T>(getArgs: GetArgs<T>): interfaces.ContainerResolution<T> {
        this.tags = [...getArgs.tags];


    private tagsChanged(tags:interfaces.Tag[]):boolean{
        let tagsDidChange = false;
        if(this.tags.length === tags.length){
            for(let i =0;i<this.tags.length;i++){
                if(this.tags[i]!== tags[i]){
                    tagsDidChange = true;
                    break;
                }
            }
        }else{
            tagsDidChange = true;
        }
        return tagsDidChange;
    }


    private _planAndResolve<T>(): (args: interfaces.NextArgs<T>) => interfaces.ContainerResolution<T> {
        return (args: interfaces.NextArgs<T>) => {
            if (!this.tagsChanged(args.tags)){
                const tags = [...args.tags];
                const firstTag = tags.shift()
                if(args.key !== firstTag![0] || args.value !== firstTag![1]){
                    if (args.key !== undefined) {
                        args.tags = [[args.key,args.value],...tags]
                    }else {
                        args.tags = tags;
                    }
                }
            }

```

luiz290788: I'll experiment with that idea.

src/interfaces/interfaces.ts
tonyhallett: and plan and and _createTags

tonyhallett: we should prevent _createTags from adding metadata with no key as well

luiz290788: > we should prevent _createTags from adding metadata with no key as well

Technically typescript is ensuring us that, right? Do we need to double check on `_createTags`?

src/container/container.ts
luiz290788: I think was able to figure it out using getters and setters. I still need to add more tests.

src/interfaces/interfaces.ts
tonyhallett: yes that's correct

src/container/container.ts
tonyhallett: The code needs to work with the NextArgs interface.  Getters and setters do not work if they are not used.  The middleware can return a different object.

I think storing on the Container is the only solution available.  
A corrected version of my idea from before.

```
private _get<T>(getArgs: GetArgs<T>): interfaces.ContainerResolution<T> {
        this.tags = getArgs.tags.map(tag => [tag[0],tag[1]]);


    private _planAndResolve<T>(): (args: interfaces.NextArgs<T>) => interfaces.ContainerResolution<T> {
        return (args: interfaces.NextArgs<T>) => {
            this._mapTagsFromKeyValue(args);

    private _tagsChanged(tags:interfaces.Tag[]):boolean{
        return this.tags.length !== tags.length ? true : this.tags.some((originalTag,i) => {
            const newTag = tags[i];
            return originalTag[0] !== newTag[0] || originalTag[1] !== newTag[1];
        });
    }
    
    private _mapTagsFromKeyValue<T>(args: interfaces.NextArgs<T>):void {
        if (!this._tagsChanged(args.tags)){
            if(this.tags.length > 0) {
                const tags = [...args.tags];
                const firstTag = tags.shift() as interfaces.Tag;
                if(args.key !== firstTag[0] || args.value !== firstTag[1]){
                    args.tags = [[args.key!,args.value],...tags];
                }
            } else if(args.key !== undefined) {
                args.tags = [[args.key,args.value]];
            }
            
        }
    }
```

tests used
```

    it("should provide key and value from first tag for old middleware", () => {
        let key:interfaces.NextArgs<unknown>["key"];
        let value:interfaces.NextArgs<unknown>["value"];
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                key = nextArgs.key;
                value = nextArgs.value;
                return next(nextArgs);
            })
        })

        container.bind("Sid").toConstantValue("Value").whenTargetMultiTagged(["tag1", 1],["tag2",2]);
        container.applyMiddleware(middleware);
        container.getTagged<string>("Sid", [["tag1", 1],["tag2",2]]);
        expect(key).to.equal("tag1");
        expect(value).to.equal(1);
    });

    it("should work with old middleware that changes value", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                if(nextArgs.key === "tag" && nextArgs.value === 1){
                    nextArgs.value = 2;
                }
                return next(nextArgs);
            })
        })

        container.applyMiddleware(middleware);
        container.bind('Sid').toConstantValue("One").whenTargetTagged("tag",1);
        container.bind('Sid').toConstantValue("Two").whenTargetTagged("tag",2);
        expect(container.getTagged<string>("Sid","tag",1)).to.equal("Two");
    });

    it("should work with old middleware that changes key", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                nextArgs.key = "new_key";
                return next(nextArgs);
            })
        })

        container.applyMiddleware(middleware);
        container.bind('Sid').toConstantValue("One").whenTargetTagged("tag",1);
        container.bind('Sid').toConstantValue("Two").whenTargetTagged("new_key",1);
        expect(container.getTagged<string>("Sid","tag",1)).to.equal("Two");
    });

    it("should work with old middleware that adds a key", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                nextArgs.key = "tag";
                nextArgs.value = 2;
                return next(nextArgs);
            })
        })

        container.applyMiddleware(middleware);
        container.bind('Sid').toConstantValue("One").whenTargetTagged("tag",1);
        container.bind('Sid').toConstantValue("Two").whenTargetTagged("tag",2);
        expect(container.get<string>("Sid")).to.equal("Two");
    });

    it("Should provide all tags for new middleware", () => {
        let tags:interfaces.Tag[]|undefined
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                tags = nextArgs.tags;
                return next(nextArgs);
            })
        })

        container.bind("Sid").toConstantValue("Value").whenTargetMultiTagged(["tag1", 1],["tag2",2]);
        container.applyMiddleware(middleware);
        container.getTagged<string>("Sid", [["tag1", 1],["tag2",2]]);
        expect(tags).to.eql([["tag1", 1],["tag2",2]]);
    });

    it("Should work with new middleware that changes key", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                const firstTag = nextArgs.tags[0];
                firstTag[0] = "new_key";
                return next(nextArgs);
            })
        })

        container.bind("Sid").toConstantValue("X").whenTargetMultiTagged(["new_key",1],["tag2",2]);
        container.applyMiddleware(middleware);
        expect(container.getTagged<string>("Sid", [["tag1", 1],["tag2",2]])).to.equal("X");
    });

    it("Should work with new middleware that changes value", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                const firstTag = nextArgs.tags[0];
                firstTag[1] = "New value";
                return next(nextArgs);
            })
        })

        container.bind("Sid").toConstantValue("X").whenTargetMultiTagged(["tag1","New value"],["tag2",2]);
        container.applyMiddleware(middleware);
        expect(container.getTagged<string>("Sid", [["tag1", 1],["tag2",2]])).to.equal("X");
    });

    it("Should work with new middleware that changes the size of tags array", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                nextArgs.tags.push(["tag2",2]);
                return next(nextArgs);
            })
        })

        container.bind("Sid").toConstantValue("X").whenTargetMultiTagged(["tag1",1],["tag2",2]);
        container.applyMiddleware(middleware);
        expect(container.getTagged<string>("Sid", [["tag1", 1]])).to.equal("X");
    });

    it("Should work with middleware returning own NextArgs", () => {
        const middleware:interfaces.Middleware = (next => {
            return (nextArgs => {
                nextArgs= {...nextArgs};
                if(nextArgs.key === "tag" && nextArgs.value === 1){
                    nextArgs.value = 2;
                }
                return next(nextArgs);
            })
        })

        container.applyMiddleware(middleware);
        container.bind('Sid').toConstantValue("One").whenTargetTagged("tag",1);
        container.bind('Sid').toConstantValue("Two").whenTargetTagged("tag",2);
        expect(container.getTagged<string>("Sid","tag",1)).to.equal("Two");
    });

```

I suggest that you look for a flaw in that and if you cannot find one then use it.

luiz290788: I though about using the container to store the tags, but wouldn't that be a problem when resolving things async?

You could have two resolutions happening at the same time and resetting the original tags before the other has finished.

I think that the probably the best way is not to memoize the set of middlewares and reuse it but create a chain of middlewares for each resolution. In that way, we can inject the original tags directly in `_planAndResolve`.

tonyhallett: > I though about using the container to store the tags, but wouldn't that be a problem when resolving things async?

Yes ignore that idea doesn't work for nested gets.

Can we use a stack of tags ?

tonyhallett: Middleware is not async.  We call a "get" method - _get sets up the tags and the args goes through the middleware until the last middleware _planAndResolve gets to fix up before resolution.  No other code can change the tags on the container before they are used.

Unless I am missing something

luiz290788: But a middleware could return anything, right?

The last middleware is the `planAndResolve` that could return a `Promise`. Let's say a middleware is returning a promise directly instead of calling the next function. Or waiting for a promise then calling the next function. We could reuse the tags from other requests

```Typescript
const myMiddleware = (next) => async (nextArgs) => {
  await somePromise()
  return next(nextArgs)
}
```

This would work because of how we are checking the result of the middleware chain, right? The next in this middleware could be `planAndResolve` and it could read the tags from another request.

tonyhallett: > But a middleware could return anything, right?

No.

```
export type Middleware = (next: Next) => Next;
export type Next = (args: NextArgs) => (any | any[]);
```

tonyhallett: If the middleware decided to kick off another "get" then putting the tags on the container would not work

tonyhallett: > If the middleware decided to kick off another "get" then putting the tags on the container would not work

The middleware does not get the container through the NextArgs although it could still have a reference to it.  If the middleware did decide to use the container ( and it was careful to not get a stack overflow ) then I think a tag stack would work.

```
private _get<T>(getArgs: GetArgs<T>): interfaces.ContainerResolution<T> {
        this.tagStack.push(getArgs.tags.map(tag => [tag[0],tag[1]]));

    private _planAndResolve<T>(): (args: interfaces.NextArgs<T>) => interfaces.ContainerResolution<T> {
        return (args: interfaces.NextArgs<T>) => {
            const tags = this.tagStack.pop()
            this._mapTagsFromKeyValue(tags!,args);

private _mapTagsFromKeyValue<T>(originalTags:interfaces.Tag[],args: interfaces.NextArgs<T>):void {
        if (!this._tagsChanged(originalTags,args.tags)){
            if(originalTags.length > 0) {
                const tags = [...args.tags];
                const firstTag = tags.shift() as interfaces.Tag;
                if(args.key !== firstTag[0] || args.value !== firstTag[1]){
                    args.tags = [[args.key!,args.value],...tags];
                }
            } else if(args.key !== undefined) {
                    args.tags = [[args.key,args.value]];
            }
        }
    }

    private _tagsChanged(originalTags:interfaces.Tag[],tags:interfaces.Tag[]):boolean{
        return originalTags.length !== tags.length ? true : originalTags.some((originalTag,i) => {
            const newTag = tags[i];
            return originalTag[0] !== newTag[0] || originalTag[1] !== newTag[1];
        });
    }

```

luiz290788: Why not building the whole chain of middlewares for each request and inject the tags in the `planAndResolve`?

tonyhallett: I assume that you mean that applyMiddleware just stores the middleware and the _get method does what applyMiddleware did before but allows fixing the original args to be consumed by planAndResolve ?

Only problem is that this is a breaking change.

luiz290788: Why is that a breaking change?

tonyhallett: Because you would be invoking the middleware for every request instead of a one time setup.

luiz290788: I see. It breaks the expectation from the user that it would be invoked only once.

tonyhallett: Perhaps then a middleware wrapping function with original nextargs and actual. Create these in applyMiddleware. Have the last in the chain no longer be planAndResolve but the fix up that will then call planAndResolve.

luiz290788: That would be a breaking change as well, wouldn't it? If the middleware creator is resolving another value and returning it we would ignore it.

tonyhallett: Will post what I think works tonight or tomorrow morning

luiz290788: I'll try to make it work with the idea of the stack.

tonyhallett: Doesn't it already ?! tagstack is just an array.

luiz290788: I've tried to implement using a version with plan and resolve outside the middleware chain. There is a problem with allowing middlewares to handle errors in resolution =(

> Doesn't it already ?! tagstack is just an array of tags.

tagStack solution is risky when a middleware code throws. We would need to be able to clean the stack. And I feel like holding state for the plan is not ideal.

tonyhallett: Will be an issue with how it determines if tags changed

tonyhallett: > feel like holding state for the plan is not ideal.

Certainly not ideal !

tonyhallett: As for the error we can just handle clear the stack and rethrow

src/container/container.ts
```diff
@@ -302,7 +304,7 @@ class Container implements interfaces.Container {
     }
 
     public applyMiddleware(...middlewares: interfaces.Middleware[]): void {
-        const initial: interfaces.Next = (this._middleware) ? this._middleware : this._planAndResolve();
+        const initial: interfaces.Next = (this._middleware) ? this._middleware : (args) => [ExecutePlanSymbol, args];
```
tonyhallett: This is an idea in progress but the end of the chain needs to return the resolved value as middleware should be able to change the resolved value.

```

        it.only("Should allow middleware to change the resolved value", () => {
            const middleware: interfaces.Middleware = (next => {
                return (nextArgs => {
                    const resolved = next(nextArgs) as string;
                    return `${resolved}+`;
                })
            })

            container.applyMiddleware(middleware);
            container.bind("Sid").toConstantValue("One");
            expect(container.get<string>("Sid")).to.equal("One+");
        })
```

src/container/container.ts
luiz290788: > As for the error we can just handle clear the stack and rethrow

I don't think that works, because at the point that we would throw the exception, the executing of the function that needs to be able to handle has already finished.

I have another idea, my last one. If that doesn't work, I'll close this pr and move this feature to `v6` and make it a breaking change removing `key` and `value` from the args.

src/container/container.ts
luiz290788: Makes sense. I'll add this test.

tonyhallett:
> Makes sense. I'll add this test.

Tests were already failing based on your last commit.  That one was just to highlight that  it did not work.

> I don't think that works, because at the point that we would throw the exception, the executing of the function that needs to be able to handle has already finished.

I assume that you mean middleware throws an exception and handles itself ?

> I have another idea, my last one. If that doesn't work, I'll close this pr and move this feature to v6 and make it a breaking change removing key and value from the args.

I look forward to your latest idea !

luiz290788:
> I assume that you mean middleware throws an exception and handles itself ?

There are tests asserting that a middleware need to be able to handle an exception that was thrown by `planAndResolve`.

tonyhallett:
> There are tests asserting that a middleware need to be able to handle an exception that was thrown by `planAndResolve`.

:thumbsup:

If only we could assume that middleware would not create a new request from the container there would be no need for a tag stack and the code I posted would be fine.

luiz290788:
> I look forward to your latest idea !

My idea does work because we are targeting `es5`, I've posted it anyway in case you know how to solve this.

tonyhallett:
> My idea does work because we are targeting es5, I've posted it anyway in case you know how to solve this.

Unfortunately this great idea will not work.
I mistakenly said that tags should not be optional on NextArgs, which would have been a breaking change.

So old middleware could return a new NextArgs object with no tags, and key and value would be lost.
`args.tags instanceof TagsArray ? args.tags.getTags(args.key, args.value) : args.tags,`

luiz290788:
> > My idea does work because we are targeting es5, I've posted it anyway in case you know how to solve this.
> 
> Unfortunately this great idea will not work.
> I mistakenly said that tags should not be optional on NextArgs, which would have been a breaking change.
> 
> So old middleware could return a new NextArgs object with no tags, and key and value would be lost.
> `args.tags instanceof TagsArray ? args.tags.getTags(args.key, args.value) : args.tags,`

I think it is time to admit defeat and focus on adding this feature to `v6` with a breaking change. What do you think?

tonyhallett:
> I think it is time to admit defeat and focus on adding this feature to v6 with a breaking change. What do you think?

I personally have had enough of it for a couple of days !

luiz290788:
> > I think it is time to admit defeat and focus on adding this feature to v6 with a breaking change. What do you think?
> 
> I personally have had enough of it for a couple of days !

Fair enough, I'll move it to v6.

tonyhallett:
Have an idea.

Although we make NextArgs.tags optional the understanding is that for new middleware it is not and that in v6 the ts  will changed accordingly.
Then

We create an internal class NextArgs.

We pass through middleware.

If is the same object that comes out in planAndResolve we can use the current logic to determine what tags should be.
If is not the same object then
a) It has no tags and so came from old middleware then we use key and value
b) It has tags and came from new middleware so we use them.


If you do not use multi tagged then there is no change to functionality.
If you want to have multi tagged then you must have compliant middleware. 

What do you think ?

tonyhallett:
```
const tags = args instanceof NextArgs ? args.getTags() : 
   args.tags ?? [[args.key,args.value]]
```

tonyhallett:
We can throw an error if is a multi tagged request and old middleware is used that changes the NextArgs object.

tonyhallett:
This will not work if old middleware creates a copy from the original that includes tags and changes key or value. Although with a getter could mark the tags array

tonyhallett:
Of course that applies to both old and new middleware. I will think again.

tonyhallett:
Perhaps add the key and value as properties of the array

luiz290788:
> Have an idea.
> 
> Although we make NextArgs.tags optional the understanding is that for new middleware it is not and that in v6 the ts will changed accordingly.
> Then
> 
> We create an internal class NextArgs.
> 
> We pass through middleware.
> 
> If is the same object that comes out in planAndResolve we can use the current logic to determine what tags should be.
> If is not the same object then
> a) It has no tags and so came from old middleware then we use key and value
> b) It has tags and came from new middleware so we use them.
> 
> If you do not use multi tagged then there is no change to functionality.
> If you want to have multi tagged then you must have compliant middleware.
> 
> What do you think ?

Makes sense. I'll try that.

tonyhallett:
I am currently on it

```
export class NextArgs<T> implements interfaces.NextArgs<T>{
    private originalTags:interfaces.Tag[]
    constructor(getArgs:GetArgs<T>){
        if(getArgs.tags === undefined){
            throw new Error("tags should be defined");
        }
        this.key = getArgs.tags[0]?.[0];
        this.value = getArgs.tags[0]?.[1];
        this.avoidConstraints = getArgs.avoidConstraints;
        this.isMultiInject = getArgs.isMultiInject;
        this.serviceIdentifier = getArgs.serviceIdentifier;
        this.originalTags = cloneTags(getArgs.tags);
        this.tags = getArgs.tags;
    }
    getTags():interfaces.Tag[]{
        return this._getTags(this.key,this.value);
    }
    private hasChanged() {
        return this.originalTags.length !== this.tags.length
          ? true
          : this.originalTags.some(
            (originalTag, i) => originalTag[0] !== this.tags[i][0] || originalTag[1] !== this.tags[i][1]);
      }

      private _getTags(key: string | symbol | number | undefined, value: unknown): interfaces.Tag[] {
        if (this.hasChanged()) {
          return this.tags;
        }
        if (this.originalTags.length > 0) {
          const tags = [...this.tags];
          const firstTag = tags.shift() as interfaces.Tag;
          if (key === undefined) {
            return tags;
          }
          if (key !== firstTag[0] || value !== firstTag[1]) {
            return [[key, value], ...tags];
          }
          return this.tags;
        } else if (key !== undefined) {
          return [[key, value]];
        }
        return []
      }

    public static getTags(nextArgs:interfaces.NextArgs<unknown>):interfaces.Tag[]{
        if(nextArgs instanceof NextArgs){
            return nextArgs.getTags();
        }
        //guaranteed old middleware
        if(nextArgs.tags===undefined){
            if(nextArgs.key !== undefined){
                return [[nextArgs.key,nextArgs.value]];
            }
            return [];
        }else{
            throw new Error("NI");
        }
    }
    avoidConstraints: boolean;
    contextInterceptor: (contexts: interfaces.Context) => interfaces.Context = (context) => context;
    isMultiInject: boolean;
    targetType = TargetTypeEnum.Variable;
    serviceIdentifier: interfaces.ServiceIdentifier<T>;
    key?: string | number | symbol | undefined;
    value?: any;
    tags: interfaces.Tag[];

}

function nextArgsTest(
            tags:interfaces.Tag[],
            middleware:(nextArgs:interfaces.NextArgs) => interfaces.NextArgs,
            expectedTags:interfaces.Tag[]){
                const nextArgs = new NextArgs({avoidConstraints:true,isMultiInject:true,serviceIdentifier:"sid",tags})
                const planAndResolveArgs = middleware(nextArgs);
                const resolvedTags = NextArgs.getTags(planAndResolveArgs);
                expect(resolvedTags.length).to.equal(expectedTags.length);
                resolvedTags.forEach((tag,i) => {
                    const expectedTag = expectedTags[i];
                    expect(tag[0]).to.equal(expectedTag[0]);
                    expect(tag[1]).to.equal(expectedTag[1]);
                })
        }

        describe.only("multi tag test", () => {
            describe("old middlware", () => {
                describe("same object", () => {
                    describe("no initial tag", () => {
                        it("should work when no tags and no change", () => {
                            nextArgsTest([],(nextArgs => {
                                return nextArgs;
                            }),[]);
                        })
                        it("should work when no tags and adds with key/value", () => {
                            nextArgsTest([],(nextArgs => {
                                nextArgs.key = "key";
                                nextArgs.value = "value";
                                return nextArgs;
                            }),[["key","value"]]);
                        });
                    })
                    describe("single tag", () => {
                        it("should work when do nothing", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                return nextArgs;
                            }),[["key","value"]]);
                        });
                        it("Should work when change key", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                nextArgs.key = "key2";
                                return nextArgs;
                            }),[["key2","value"]]);
                        })
                        it("Should work when change value", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                nextArgs.value = "value2";
                                return nextArgs;
                            }),[["key","value2"]]);
                        })
                    });
                    describe("multi tag", () => {
                        //
                    });

                });
                describe("different object", () => {
                    describe("but does not copy tags", () => {
                        describe("no tags", () => {
                            it("should work when no tags and no change", () => {
                                nextArgsTest([],(nextArgs => {
                                    return nextArgs;
                                }),[]);
                            })
                            it("should work when no tags and adds with key/value", () => {
                                nextArgsTest([],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    copy.key = "key";
                                    copy.value = "value";
                                    return copy;
                                }),[["key","value"]]);
                            });
                        });
                        describe("single tag", () => {
                            it("should work if does nothing", () => {
                                nextArgsTest([["key","value"]],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    return copy;
                                }),[["key","value"]]);
                            })
                            it("should work when tag and changes key", () => {
                                nextArgsTest([["key","value"]],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    copy.key = "key2";
                                    return copy;
                                }),[["key2","value"]]);
                            });
                            it("should work when tag and changes value", () => {
                                nextArgsTest([["key","value"]],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    copy.value = "value2";
                                    return copy;
                                }),[["key","value2"]]);
                            });
                        })
                        describe("multi tags", () => {
                            //this loses additional tags - then an exception can be thrown
                        })
                    });
                    //************************************************* */
                    describe("but does copy tags", () => {
                        //
                    });
                })
            });

            describe("new middleware", () => {
                describe("same object", () => {
                    describe("no initial tag", () => {
                        it("should work when no tags and no change", () => {
                            nextArgsTest([],(nextArgs => {
                                return nextArgs;
                            }),[]);
                        })
                        it("should work when no tags and adds to tags array", () => {
                            nextArgsTest([],(nextArgs => {
                                nextArgs.tags?.push(["key","value"])
                                return nextArgs;
                            }),[["key","value"]]);
                        });

                        it("should work when no tags and sets a new tags array", () => {
                            nextArgsTest([],(nextArgs => {
                                nextArgs.tags = [["key","value"]];
                                return nextArgs;
                            }),[["key","value"]]);
                        });
                    });
                    describe("initial multitags")
                })

                // different array
                describe("different object", () => {
                    //....
                })
            })
        })

```

The code will change as I progress

tonyhallett:
I think this works

```
type MarkedTags = interfaces.Tag[] & {original:true,key:string|number|symbol|undefined,value:any}
export class NextArgs<T> implements interfaces.NextArgs<T>{
    private originalTags:interfaces.Tag[]
    constructor(getArgs:GetArgs<T>){
        if(getArgs.tags === undefined){
            throw new Error("tags should be defined");
        }
        this.key = getArgs.tags[0]?.[0];
        this.value = getArgs.tags[0]?.[1];
        this.avoidConstraints = getArgs.avoidConstraints;
        this.isMultiInject = getArgs.isMultiInject;
        this.serviceIdentifier = getArgs.serviceIdentifier;
        this.originalTags = cloneTags(getArgs.tags);
        this.tags = getArgs.tags;
        const markedTags:MarkedTags = (this.tags as any);
        markedTags.original = true;
        markedTags.key = this.key;
        markedTags.value = this.value;
    }
    getTags():interfaces.Tag[]{
        return this._getTags(this.key,this.value);
    }
    private hasChanged() {
        return this.originalTags.length !== this.tags.length
          ? true
          : this.originalTags.some(
            (originalTag, i) => originalTag[0] !== this.tags[i][0] || originalTag[1] !== this.tags[i][1]);
      }

      private _getTags(key: string | symbol | number | undefined, value: unknown): interfaces.Tag[] {
        if (this.hasChanged()) {
          return this.tags;
        }
        if (this.originalTags.length > 0) {
          const tags = [...this.tags];
          const firstTag = tags.shift() as interfaces.Tag;
          if (key === undefined) {
            return tags;
          }
          if (key !== firstTag[0] || value !== firstTag[1]) {
            return [[key, value], ...tags];
          }
          return this.tags;
        } else if (key !== undefined) {
          return [[key, value]];
        }
        return []
      }

    private static isMarkedTags(tags:interfaces.Tag[]): tags is MarkedTags {
        return (tags as MarkedTags).original;
    }
    public static getTags(nextArgs:interfaces.NextArgs<unknown>):interfaces.Tag[]{
        if(nextArgs instanceof NextArgs){
            return nextArgs.getTags();
        }
        const tags = nextArgs.tags;
        //guaranteed old middleware
        if(tags===undefined){
            if(nextArgs.key !== undefined){
                return [[nextArgs.key,nextArgs.value]];
            }
            return [];
        }
        if(this.isMarkedTags(tags)){
            const originalKey = tags.key;
            const originalValue = tags.value;
            if(nextArgs.key !== originalKey || nextArgs.value !== originalValue){
                if(nextArgs.key !== undefined){
                    return [[nextArgs.key,nextArgs.value]];
                }
                return [];
            }else{
                return tags;
            }
        }else{
            return tags;
        }
    }
    avoidConstraints: boolean;
    contextInterceptor: (contexts: interfaces.Context) => interfaces.Context = (context) => context;
    isMultiInject: boolean;
    targetType = TargetTypeEnum.Variable;
    serviceIdentifier: interfaces.ServiceIdentifier<T>;
    key?: string | number | symbol | undefined;
    value?: any;
    tags: interfaces.Tag[];

}
describe.only("multi tag test", () => {
            describe("old middlware", () => {
                describe("same object", () => {
                    describe("no initial tag", () => {
                        it("should work when no tags and no change", () => {
                            nextArgsTest([],(nextArgs => {
                                return nextArgs;
                            }),[]);
                        })
                        it("should work when no tags and adds with key/value", () => {
                            nextArgsTest([],(nextArgs => {
                                nextArgs.key = "key";
                                nextArgs.value = "value";
                                return nextArgs;
                            }),[["key","value"]]);
                        });
                    })
                    describe("single tag", () => {
                        it("should work when do nothing", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                return nextArgs;
                            }),[["key","value"]]);
                        });
                        it("Should work when change key", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                nextArgs.key = "key2";
                                return nextArgs;
                            }),[["key2","value"]]);
                        })
                        it("Should work when change value", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                nextArgs.value = "value2";
                                return nextArgs;
                            }),[["key","value2"]]);
                        })
                    });
                    describe("multi tag", () => {
                        it("should apply key value change", () => {
                            nextArgsTest([["key","value"],["key2","value2"]],(nextArgs => {
                                nextArgs.key = "new key";
                                nextArgs.value = "new value";
                                return nextArgs;
                            }),[["new key","new value"],["key2","value2"]]);
                        })
                    });

                });
                describe("different object", () => {
                    describe("but does not copy tags", () => {
                        describe("no tags", () => {
                            it("should work when no tags and no change", () => {
                                nextArgsTest([],(nextArgs => {
                                    return nextArgs;
                                }),[]);
                            })
                            it("should work when no tags and adds with key/value", () => {
                                nextArgsTest([],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    copy.key = "key";
                                    copy.value = "value";
                                    return copy;
                                }),[["key","value"]]);
                            });
                        });
                        describe("single tag", () => {
                            it("should work if does nothing", () => {
                                nextArgsTest([["key","value"]],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    return copy;
                                }),[["key","value"]]);
                            })
                            it("should work when tag and changes key", () => {
                                nextArgsTest([["key","value"]],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    copy.key = "key2";
                                    return copy;
                                }),[["key2","value"]]);
                            });
                            it("should work when tag and changes value", () => {
                                nextArgsTest([["key","value"]],(nextArgs => {
                                    const {tags, ...copy} = {...nextArgs};
                                    copy.value = "value2";
                                    return copy;
                                }),[["key","value2"]]);
                            });
                        })
                        describe("multi tags", () => {
                            //this loses additional tags - then an exception can be thrown
                        })
                    });
                    describe("but does copy tags", () => {
                        it("should use the key change", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                const copy = {...nextArgs};
                                copy.key = "new key";
                                return copy;
                            }),[["new key","value"]]);
                        });

                        it("should use the value change", () => {
                            nextArgsTest([["key","value"]],(nextArgs => {
                                const copy = {...nextArgs};
                                copy.value = "new value";
                                return copy;
                            }),[["key","new value"]]);
                        })
                    });
                })
            });

            describe("new middleware", () => {
                describe("same object", () => {
                    describe("no initial tag", () => {
                        it("should work when no tags and no change", () => {
                            nextArgsTest([],(nextArgs => {
                                return nextArgs;
                            }),[]);
                        })
                        it("should work when no tags and adds to tags array", () => {
                            nextArgsTest([],(nextArgs => {
                                nextArgs.tags?.push(["key","value"])
                                return nextArgs;
                            }),[["key","value"]]);
                        });

                        it("should work when no tags and sets a new tags array", () => {
                            nextArgsTest([],(nextArgs => {
                                nextArgs.tags = [["key","value"]];
                                return nextArgs;
                            }),[["key","value"]]);
                        });
                    });
                    describe("initial multitags",() => {
                        it("should work when multi tags and no change", () => {
                            nextArgsTest([["key1","value1"],["key2","value2"]],(nextArgs => {
                                return nextArgs;
                            }),[["key1","value1"],["key2","value2"]]);
                        })

                        it("should work when multi tags and add tag", () => {
                            nextArgsTest([["key1","value1"],["key2","value2"]],(nextArgs => {
                                nextArgs.tags?.push(["key3","value3"]);
                                return nextArgs;
                            }),[["key1","value1"],["key2","value2"],["key3","value3"]]);
                        })

                        it("should work when multi tags and sets a new tags array", () => {
                            nextArgsTest([["key1","value"],["key2","value2"]],(nextArgs => {
                                nextArgs.tags = [["key3","value3"],["key4","value4"]];
                                return nextArgs;
                            }),[["key3","value3"],["key4","value4"]]);
                        })
                    })
                })

                describe("different object", () => {
                    describe("same tags array", () => {
                        it("should notice the change", () => {
                            nextArgsTest([["key1","value1"],["key2","value2"]],(nextArgs => {
                                const copy = {...nextArgs};
                                copy.tags?.push(["key3","value3"])
                                return copy;
                            }),[["key1","value1"],["key2","value2"],["key3","value3"]]);
                        })
                    })
                    describe("different tags array", () => {
                        it("should use these tags", () => {
                            nextArgsTest([["key1","value1"],["key2","value2"]],(nextArgs => {
                                const copy = {...nextArgs};
                                copy.tags = [["key3","value3"],["key4","value4"]];
                                return copy;
                            }),[["key3","value3"],["key4","value4"]]);
                        });
                    })
                })
            })
        })
```

Can you see a flaw ?

luiz290788:
> Can you see a flaw ?

I cannot have a look at that right now. I'll look it carefully this night.

tonyhallett:
A tidier version of NextArgs

```
type Key = string | symbol | number | undefined;

type MarkedTags = interfaces.Tag[] & {original:true,key:Key,value:any}

export class NextArgs<T> implements interfaces.NextArgs<T>{
    private _originalTags:interfaces.Tag[]

    avoidConstraints: boolean;
    contextInterceptor: (contexts: interfaces.Context) => interfaces.Context = (context) => context;
    isMultiInject: boolean;
    targetType = TargetTypeEnum.Variable;
    serviceIdentifier: interfaces.ServiceIdentifier<T>;
    key?: Key;
    value?: any;
    tags: interfaces.Tag[];

    constructor(getArgs:GetArgs<T>){
        if(getArgs.tags === undefined){
            throw new Error("tags should be defined");
        }
        this.key = getArgs.tags[0]?.[0];
        this.value = getArgs.tags[0]?.[1];
        this.avoidConstraints = getArgs.avoidConstraints;
        this.isMultiInject = getArgs.isMultiInject;
        this.serviceIdentifier = getArgs.serviceIdentifier;
        this._setTags(getArgs.tags);
    }
    private _cloneTags(tags: interfaces.Tag[]): interfaces.Tag[] {
        return tags.map(tag => [tag[0], tag[1]])
    }

    private _setTags(tags:interfaces.Tag[]){
        this._originalTags = this._cloneTags(tags);
        this.tags = tags;
        this._markTags();
    }

    private _markTags(){
        const markedTags = this.tags as MarkedTags;
        markedTags.original = true;
        markedTags.key = this.key;
        markedTags.value = this.value;
    }

    private hasChanged() {
        return this._originalTags.length !== this.tags.length
            ? true
            : this._originalTags.some(
                (originalTag, i) => originalTag[0] !== this.tags[i][0] || originalTag[1] !== this.tags[i][1]
            );
    }

    private _getTags(key: Key, value: unknown): interfaces.Tag[] {
        if (this.hasChanged()) {
            return this.tags;
        }
        return NextArgs._fixUpTags(key,value,this.tags)
    }

    getTags():interfaces.Tag[]{
        return this._getTags(this.key,this.value);
    }

    private static _fixUpTags(key: Key, value: unknown,tags:interfaces.Tag[]): interfaces.Tag[]{
        if (tags.length > 0) {
            const remainingTags = [...tags];
            const firstTag = remainingTags.shift() as interfaces.Tag;
            if (key === undefined) {
              return remainingTags;
            }
            if (key !== firstTag[0] || value !== firstTag[1]) {
              return [[key, value], ...remainingTags];
            }
            return tags;
        } else if (key !== undefined) {
            return [[key, value]];
        }
        return []
    }

    private static _isMarkedTags(tags:interfaces.Tag[]): tags is MarkedTags {
        return (tags as MarkedTags).original;
    }

    private static _getTagsIfGuaranteedOldMiddleware(tags:interfaces.Tag[]|undefined,key:Key,value:unknown):interfaces.Tag[] | null {
        if(tags===undefined){
            return key === undefined ? [] : [[key,value]];
        }
        return null;
    }

    private static _getTagsBasedOnMarks(tags:interfaces.Tag[],key:Key,value:unknown):interfaces.Tag[] {
        if(NextArgs._isMarkedTags(tags)){
            const originalKey = tags.key;
            const originalValue = tags.value;
            if(key !== originalKey || value !== originalValue){
                tags = NextArgs._fixUpTags(key,value,tags);
            }
        }
        return tags;
    }

    private static _getTagsFromChanged(tags:interfaces.Tag[]|undefined,key:Key,value:unknown):interfaces.Tag[] {
        return NextArgs._getTagsIfGuaranteedOldMiddleware(tags,key,value) ??
            NextArgs._getTagsBasedOnMarks(tags!,key,value);
    }

    public static getTags(nextArgs:interfaces.NextArgs<unknown>):interfaces.Tag[]{
        if(nextArgs instanceof NextArgs){
            return nextArgs.getTags();
        }
        return NextArgs._getTagsFromChanged(nextArgs.tags,nextArgs.key,nextArgs.value);
    }

}
```

and of course remember to use them in planAndResolve

```
private _planAndResolve<T>(): (args: interfaces.NextArgs<T>) => interfaces.ContainerResolution<T> {
        return (args: interfaces.NextArgs<T>) => {

            // create a plan
            let context = plan(
                this._metadataReader,
                this,
                args.isMultiInject,
                args.targetType,
                args.serviceIdentifier,
                args.tags = NextArgs.getTags(args),
                args.avoidConstraints
            );
```

tonyhallett:
I think that it can be made simpler.

```
export class NextArgs<T> implements interfaces.NextArgs<T>{
    avoidConstraints: boolean;
    contextInterceptor: (contexts: interfaces.Context) => interfaces.Context = (context) => context;
    isMultiInject: boolean;
    targetType = TargetTypeEnum.Variable;
    serviceIdentifier: interfaces.ServiceIdentifier<T>;
    key?: Key;
    value?: any;
    tags: interfaces.Tag[];

    constructor(getArgs:GetArgs<T>){
        if(getArgs.tags === undefined){
            throw new Error("tags should be defined");
        }
        this.key = getArgs.tags[0]?.[0];
        this.value = getArgs.tags[0]?.[1];
        this.avoidConstraints = getArgs.avoidConstraints;
        this.isMultiInject = getArgs.isMultiInject;
        this.serviceIdentifier = getArgs.serviceIdentifier;
        this._setTags(getArgs.tags);
    }

    private _setTags(tags:interfaces.Tag[]){
        this.tags = tags;
        this._markTags();
    }

    private _markTags(){
        const markedTags = this.tags as MarkedTags;
        markedTags.original = true;
        markedTags.key = this.key;
        markedTags.value = this.value;
    }

    private static _fixUpTags(key: Key, value: unknown,tags:interfaces.Tag[]): interfaces.Tag[]{
        if (tags.length > 0) {
            const remainingTags = [...tags];
            const firstTag = remainingTags.shift() as interfaces.Tag;
            if (key === undefined) {
              return remainingTags;
            }
            if (key !== firstTag[0] || value !== firstTag[1]) {
              return [[key, value], ...remainingTags];
            }
            return tags;
        } else if (key !== undefined) {
            return [[key, value]];
        }
        return []
    }

    private static _isMarkedTags(tags:interfaces.Tag[]): tags is MarkedTags {
        return (tags as MarkedTags).original;
    }

    private static _getTagsIfGuaranteedOldMiddleware(tags:interfaces.Tag[]|undefined,key:Key,value:unknown):interfaces.Tag[] | null {
        if(tags===undefined){
            return key === undefined ? [] : [[key,value]];
        }
        return null;
    }

    private static _getTagsBasedOnMarks(tags:interfaces.Tag[],key:Key,value:unknown):interfaces.Tag[] {
        if(NextArgs._isMarkedTags(tags)){
            const originalKey = tags.key;
            const originalValue = tags.value;
            if(key !== originalKey || value !== originalValue){
                tags = NextArgs._fixUpTags(key,value,tags);
            }
        }
        return tags;
    }

    private static _getTags(tags:interfaces.Tag[]|undefined,key:Key,value:unknown):interfaces.Tag[] {
        return NextArgs._getTagsIfGuaranteedOldMiddleware(tags,key,value) ??
            NextArgs._getTagsBasedOnMarks(tags!,key,value);
    }

    public static getTags(nextArgs:interfaces.NextArgs<unknown>):interfaces.Tag[]{
        return NextArgs._getTags(nextArgs.tags,nextArgs.key,nextArgs.value);
    }

}
```

luiz290788:
@tonyhallett  So, I've applied your changes. It seems good but I think we have a problem when we decide to mix middlewares that use `tag` and `value` with middlewares that use `tags`.

So, I'm inclined to delay this change to the next major version. I'll work on a PR for that.

tonyhallett:
> It seems good but I think we have a problem when we decide to mix middlewares that use tag and value with middlewares that use tags.

If we mix middleware then old middleware could change key and value and this would not be reflected in tags.

If we state that until vN there is a condition on middleware then at least the functionality is available now.

I say vN because there is no guarantee that this feature will be included in the next release.

We just need to carefully state this condition.

luiz290788:
> If we state that until vN there is a condition on middleware then at least the functionality is available now.

What is the condition? that you shouldn't change `key` and `value` in a middleware and change `tags` in another middleware?

tonyhallett:
> What is the condition? that you shouldn't change key and value in a middleware and change tags in another middleware?

If all the applied middleware ( mixture of new and old ) just read next args then all good.  Is this the most likely usage of middleware ?

If using only old middleware then it can change key/value but tags other than the first could be lost ( if it creates a new NextArgs  - see explanation)

If using only new middleware then changes to tags will flow through as expected.

If combining old and new middleware and tags are changed through either mechanism.
Describe what we are doing.
NextArgs begins with tags and key and value being the key and value of the first tag ( if there is one ).
Tags has the key and value attached so we can try to apply old middleware changes at the end.

After all middleware has determined the final NextArgs.
If tags is not present ( old middleware created new NextArgs ) then key/value becomes tags.
If the marked tags is not present ( new middleware ) we use those tags.
If the marked tags are present and key or value have changed then key/value determine the first tag ( and the remainder are applied )

To be aware of
Old middleware creating a new NextArgs without object spread leading to no tags for new middleware along the chain.
Old middleware changing key/value to new objects will not be represented on tags for new middleware along the chain.
New middleware changing the first tag will not be represented in key/value for old middleware along the chain.

If new or old middleware change a tag value object by changing its state then it will be reflected.

> Old middleware changing key/value to new objects will not be represented on tags for new middleware along the chain.

Perhaps with a setter we could improve this.

> New middleware changing the first tag will not be represented in key/value for old middleware along the chain.

Again perhaps we could try and limit this exposure.

tonyhallett:
> Old middleware changing key/value to new objects will not be represented on tags for new middleware along the chain.

It would be simple with a setter to update the first tag

By doing this we can have a scenario where 
old middleware can change key/value -> next new middleware can see the change and change again and take precedence as the key/value setter can update the tags array and the attached key/value.

As for

> If the marked tags is not present ( new middleware ) we use those tags.

Inform devs not to do it.

tonyhallett:
Final thought

> New middleware changing the first tag will not be represented in key/value for old middleware along the chain.

If we change NextArgs.tags to not be a Tag[] but a Tags class we could be aware of changes to key/value for the first tag or the first tag itself and update NextArgs.key and NextArgs.value.

Not as dev friendly as working with an array but combined with a key/value setter we could ensure that the two are always aligned and the only consideration for devs is of old middleware that returned new NextArgs without object spread.


Your thoughts ?

PR #1383, opened 2021-10-07
Update typescript compiler options  fix

jkhmtx:
Fixes a Node 9/10 specific issue with the way that circular dependency errors are managed. The issue relates to an inability to access `rootRequest` on `context.plan` in the error management code because `context.plan` is `undefined`, which throws an unexpected error.


jkhmtx:
Going to base this against `master` for now just so that the CI will run, and then base it against the proper TS update feature branch.

jkhmtx:
The last three commits are meant to resolve the outstanding CodeClimate issues that have been introduced in this PR.

PR #1407, opened 2021-12-16
prepare typescript 4.5

PodaruDragos:
Prepare for TS 4.5


.github/workflows/ci.yml
```diff
@@ -91,4 +91,4 @@ jobs:
         uses: codecov/codecov-action@f32b3a3741e1053eb607407145bc9619351dc93b # renovate: tag=v2
         with:
           directory: coverage/
-          fail_ci_if_error: true
+          fail_ci_if_error: true
```
dcavanagh, line 94: missing new line

src/bindings/binding_count.ts
```diff
@@ -1,7 +1,5 @@
-const BindingCount = {
+export const BindingCount = {
   MultipleBindingsAvailable: 2,
   NoBindingsAvailable: 0,
   OnlyOneBindingAvailable: 1
-};
-
-export { BindingCount };
+};
```
dcavanagh, line 5: missing new line

src/utils/exceptions.ts
```diff
@@ -1,9 +1,9 @@
-import * as ERROR_MSGS from "../constants/error_msgs";
+import * as ERROR_MSGS from '../constants/error_msgs';
 
-export function isStackOverflowExeption(error: Error) {
+export function isStackOverflowExeption(error: unknown | Error): error is RangeError {
```
jkhmtx: Since `unknown` is a top type, `unknown | Error` is the equivalent of simply `unknown` to the compiler.

Lovely to see the usage of the `is RangeError` type predicate though :+1: 
```suggestion
export function isStackOverflowExeption(error: unknown): error is RangeError {
```

jkhmtx: I have noticed this situation of using `SomeType | unknown` at least one other time, and I will try to find it, but I know I may miss it.

src/syntax/binding_to_syntax.ts
```diff
@@ -6,7 +6,7 @@ import { BindingWhenOnSyntax } from "./binding_when_on_syntax";
 
 class BindingToSyntax<T> implements interfaces.BindingToSyntax<T> {
 
-  private _binding: interfaces.Binding<T>;
+  public _binding: interfaces.Binding<T>;
```
jkhmtx: For this, I think we should keep this member `private` and use the same tricks that are in use elsewhere to make it accessible with an `as unknown as { _binding: interfaces.Binding<T> }`, but only _for now_ :). Please see my following comment.

jkhmtx: I think that this is uncovering a smell.

I have noticed in the past that this member is `private`, but a type assertion is used in order to allow it to be accessed from a public location. This is a pattern that can be observed with other members, not just this one (for instance, `_bindingDictionary`).

A few places I have noticed this are among test files:

[`test/container/container.test.ts`](https://github.com/inversify/InversifyJS/blob/master/test/container/container.test.ts#L608)
[`test/syntax/binding_when_on_syntax.test.ts`](https://github.com/inversify/InversifyJS/blob/master/test/syntax/binding_when_on_syntax.test.ts#L32)

Another place is in [`src/planning/planner.ts`](https://github.com/inversify/InversifyJS/blob/master/src/planning/planner.ts#L21).

The reality of this smell is that there is a need for two forms of the containing types (in this case, I am referring to `BindingToSyntax` since this is where we have concentrated this specific example):

1. A _presentational_ version which maintains a correct API for the consumers of the package, and does not expose unnecessary members like `_binding`.
2. An _internal_ version which allows a more exposed API for the purposes of testing and proper layer integration with `planner` and friends.

I think that a good way of accomplishing this which resolves the smell is to rename the class with a `_` prefix, as in `BindingToSyntax => _BindingToSyntax`, etc. This means that the current definition of this class will become the internal form of the class. Then, any members which are meant to be public may be accomplished in the following manner:

```typescript
// Please ignore the lack of generic type vars, etc.,
// as this is only meant to be an example
class BindingToSyntax {
  constructor(
    private _internal = new _BindingToSyntax(...args)
  ) {}

  get somePublicMemberFromOriginalImplementation() {
    return this._internal.somePublicMemberFromOriginalImplementation;
  }
}
```

etc.

Finally, with a presentational/internal pattern like this, we are free to fix the original problem for those classes which have the behavior of needing a private member exposed for testing, or for use in `planner` or similar. We can simply make the members that needed this special access despite being `private` members into `public` ones instead.

src/planning/reflection_utils.ts
```diff
@@ -76,8 +76,8 @@ function getConstructorArgsAsTarget(
   // Take types to be injected from user-generated metadata
   // if not available use compiler-generated metadata
   let serviceIdentifier = serviceIdentifiers[index];
-  const injectIdentifier = (metadata.inject || metadata.multiInject);
-  serviceIdentifier = (injectIdentifier) ? (injectIdentifier) : serviceIdentifier;
+  const injectIdentifier = metadata.inject || metadata.multiInject;
+  serviceIdentifier = (injectIdentifier ? injectIdentifier : serviceIdentifier) as interfaces.ServiceIdentifier<unknown> | undefined;
```
jkhmtx, line 80: I know it was just simple type annotations being added here, but since `injectIdentifier` isn't used outside of the assignment on line 80, it would be great if someday we could rewrite this in the following manner:

```typescript
let serviceIdentifier = metadata.inject ?? metadata.multiInject ?? serviceIdentifiers[index];
```

Ideally this would avoid needing an explicit cast to `interfaces.ServiceIdentifier<unknown> | undefined`, since (I hope) we have `strictNullChecks` enabled.

This would be better enabled by making the return type of `formatTargetMetadata` use `interfaces.ServiceIdentifier<unknown>` for the `inject` and `multiInject` properties.

src/interfaces/interfaces.ts
```diff
@@ -90,7 +92,7 @@ namespace interfaces {
     factory: FactoryTypeFunction | null
   };
 
-  export type Provider<T> = (...args: any[]) => (((...args: any[]) => Promise<T>) | Promise<T>);
+  export type Provider<T> = (...args: never[]) => (((...args: never[]) => Promise<T>) | Promise<T>);
```
jkhmtx: I would highly recommend leaving this type as `any`. `never` is not really a good candidate for this use case.

In my mind, `any` is acceptable for defining function args with unknown arity.

src/interfaces/interfaces.ts
```diff
@@ -40,6 +40,8 @@ namespace interfaces {
 
   export type Newable<T> = new (...args: never[]) => T;
 
+  export type Instance<T> = T & Record<string, () => void>;
```
jkhmtx, line 43: Interesting choice here. I think there may be a better name for this type as it relates to the context it is used in.

1. I like the idea of encapsulating this type with `type` keyword.
2. Calling it `Instance` feels a little undercooked since the right hand side of the assignment does not really imply to me that this type is exlusively `Instance`-like. Does it seem possible that we could infer a better name from its usage?
3. Should we consider whether this should be exported publicly from the package? We tend to put all interfaces in this module, which makes them available to consumers of the package, but thinking of proper encapsulation, does a consumer of the package care about this type? Is there a better/closer place to put?

I'm personally ready to challenge the assumption that using `interfaces.ts` for all interfaces and types is the correct pattern, but it's certainly not in scope for this one issue :) Maybe we should keep it here for now and thinking about breaking apart `interfaces` in a future breaking changes release?

src/bindings/binding_count.ts
```diff
@@ -1,7 +1,5 @@
-const BindingCount = {
+export const BindingCount = {
```
jkhmtx, line 1: I like this a lot better. Maybe in the future this is better as an `enum`?

jkhmtx (changes requested):
Thank you @PodaruDragos for your beautiful work in this PR :pray: 

I have a few things that I would like to see changes about, and I am happy to discuss them more:

1. Please reconsider the usage of `never` when defining a function type of unknown arity. (this tends to introduce complications with using the type and interacting with other types that depend on `...unknown[]`).
2. Please reconsider making a member `public` without addressing the underlying code smell (please see my comment for a writeup about this).

src/syntax/binding_to_syntax.ts
jkhmtx: If you agree @PodaruDragos then I will spin the above comment off as an issue for us to work on at some time.

PodaruDragos: totally agree with this. Do you also agree that we should just move everything to actual `#private` fields since you know they are really `private` ?

package.json
```diff
@@ -1,48 +1,9 @@
 {
-  "name": "inversify",
```
dcavanagh, line 2: Not sure if we need this reorganized.

src/interfaces/interfaces.ts
dcavanagh: @jakehamtexas I agree about the interfaces. interfaces.ts should really only export the interfaces we want to see in the public api.

src/syntax/binding_to_syntax.ts
dcavanagh: I think it is worth taking a look at cleaning this up

src/utils/exceptions.ts
dcavanagh: Would `Error | unknown` work better?

test/annotation/post_construct.test.ts
```diff
@@ -50,10 +50,10 @@ describe("@postConstruct", () => {
...
 
-    decorate(postConstruct(), VanillaJSWarrior.prototype, "testMethod");
+    decorate(postConstruct(), VanillaJSWarrior.prototype, 'testMethod');
 
     const metadata: Metadata = Reflect.getMetadata(METADATA_KEY.POST_CONSTRUCT, VanillaJSWarrior);
-    expect(metadata.value).to.be.equal("testMethod");
+    expect(metadata.value).to.be.equal('testMethod');
   });
 
 });
```
dcavanagh, line 59: new line missing

test/bindings/binding.test.ts
```diff
@@ -1,19 +1,19 @@
...
     const fooBinding = new Binding<Stubs.FooInterface>(fooIdentifier, BindingScopeEnum.Transient);
     expect(fooBinding.serviceIdentifier).eql(fooIdentifier);
     expect(fooBinding.implementationType).eql(null);
     expect(fooBinding.cache).eql(null);
     expect(fooBinding.scope).eql(BindingScopeEnum.Transient);
-    expect(fooBinding.id).to.be.a("number");
+    expect(fooBinding.id).to.be.a('number');
   });
 
 });
```
dcavanagh, line 19: missing new line

test/bugs/bugs.test.ts
```diff
@@ -127,20 +127,20 @@ describe("Bugs", () => {
...
+      Rank: 'Rank',
+      Weapon: 'Weapon'
     };
 
     @injectable()
     class SamuraiMaster extends Warrior {
       public weapon: Weapon;
       public constructor(
-        @inject(TYPES.Rank) @named("master") public rank: string,
+        @inject(TYPES.Rank) @named('master') public override rank: string,
```
dcavanagh, line 143: ~~I just read the `override` docs and it made me think that we should enable `noImplicitOverride`~~
saw that you added it

test/bugs/issue_549.test.ts
```diff
@@ -52,23 +52,23 @@ describe("Issue 549", () => {
...
-
-      const expectedErrorA = ERROR_MSGS.CIRCULAR_DEPENDENCY_IN_FACTORY("toDynamicValue", TYPE.ADynamicValue.toString());
-      const expectedErrorB = ERROR_MSGS.CIRCULAR_DEPENDENCY_IN_FACTORY("toDynamicValue", TYPE.BDynamicValue.toString());
-      const matchesErrorA = e.message.indexOf(expectedErrorA) !== -1;
-      const matchesErrorB = e.message.indexOf(expectedErrorB) !== -1;
+      const localError = e as Error;
+      const expectedErrorA = ERROR_MSGS.CIRCULAR_DEPENDENCY_IN_FACTORY('toDynamicValue', TYPE.ADynamicValue.toString());
+      const expectedErrorB = ERROR_MSGS.CIRCULAR_DEPENDENCY_IN_FACTORY('toDynamicValue', TYPE.BDynamicValue.toString());
+      const matchesErrorA = localError.message.indexOf(expectedErrorA) !== -1;
+      const matchesErrorB = localError.message.indexOf(expectedErrorB) !== -1;
```
dcavanagh, line 62: Not sure if this is better or not
```suggestion
      const matchesErrorA = (e as Error).message.indexOf(expectedErrorA) !== -1;
      const matchesErrorB = (e as Error).message.indexOf(expectedErrorB) !== -1;
```

test/inversify.test.ts
```diff
@@ -1,15 +1,11 @@
...
-import { interfaces } from "../src/interfaces/interfaces";
-import {
-  Container, ContainerModule, decorate, inject,
-  injectable, LazyServiceIdentifer, multiInject, named, tagged, targetName,
-  typeConstraint, unmanaged
-} from "../src/inversify";
+import { expect } from 'chai';
+import * as ERROR_MSGS from '../src/constants/error_msgs';
+import { interfaces } from '../src/interfaces/interfaces';
+import { Container, ContainerModule, decorate, inject, injectable, LazyServiceIdentifer, multiInject, named, tagged, targetName, typeConstraint, unmanaged } from '../src/inversify';
```
dcavanagh, line 4:
```suggestion
import { 
  Container, 
  ContainerModule, 
  decorate, 
  inject, 
  injectable, 
  LazyServiceIdentifer, 
  multiInject, 
  named, 
  tagged, 
  targetName, 
  typeConstraint, 
  unmanaged 
} from '../src/inversify';
```

test/planning/context.test.ts
```diff
@@ -40,7 +40,7 @@ describe("Context", () => {
     const plan = new Plan(context, ninjaRequest);
     context.addPlan(plan);
 
-    expect(context.plan.rootRequest.serviceIdentifier).eql("Ninja");
+    expect(context.plan.rootRequest.serviceIdentifier).eql('Ninja');
   });
 
 });
```
dcavanagh, line 46: missing new line

test/utils/binding_utils.test.ts
```diff
@@ -1,11 +1,11 @@
...
-    const binding = new Binding("", "Singleton");
-    binding.type = "Instance";
-    expect(() => getFactoryDetails(binding)).to.throw("Unexpected factory type Instance");
+describe('getFactoryDetails', () => {
+  it('should thrown an exception non factory binding.type', () => {
+    const binding = new Binding('', 'Singleton');
+    binding.type = 'Instance';
+    expect(() => getFactoryDetails(binding)).to.throw('Unexpected factory type Instance');
   });
 })
```
dcavanagh, line 11: missing new line

test/utils/id.test.ts
```diff
@@ -1,11 +1,11 @@
...
+describe('ID', () => {
 
-  it("Should be able to generate an id", () => {
+  it('Should be able to generate an id', () => {
     const id1 = id();
-    expect(id1).to.be.a("number");
+    expect(id1).to.be.a('number');
   });
 
 });
```
dcavanagh, line 11: missing newline

test/utils/serialization.test.ts
```diff
@@ -1,43 +1,43 @@
...
+  it('Should extract symbol description', () => {
+    const symbolWithDescription = Symbol('description');
+    expect(getSymbolDescription(symbolWithDescription)).to.equal('description');
 
     const symbolWithoutDescription = Symbol();
-    expect(getSymbolDescription(symbolWithoutDescription)).to.equal("");
+    expect(getSymbolDescription(symbolWithoutDescription)).to.equal('');
   });
 
 });
```
dcavanagh, line 43: missing new line

tslint.json
```diff
@@ -20,7 +20,10 @@
     "max-file-line-count": false,
     "max-line-length": [
       true,
-      140
```
dcavanagh, line 23: I think I would rather have the imports have line breaks after 3

dcavanagh (changes requested):
Thanks for all the work. Looking good and generating some discussion!

src/utils/exceptions.ts
jkhmtx: No, unfortunately it is the same no matter the order @dcavanagh 

Any type union with `unknown` will simply be `unknown` without any of the other union members

src/syntax/binding_to_syntax.ts
jkhmtx: @PodaruDragos I think we should definitely do `#private` type of vars in the future, for now let's not increase the scope of this PR

test/annotation/post_construct.test.ts
jkhmtx: Why is this a problem? Let's make the linting and formatting automatic in a different scoped PR, unless there is a pressing reason to make this (and the other) formatting correction(s)

test/bugs/issue_549.test.ts
jkhmtx: I think the existing version has less redundancy, I would recommend against this suggestion

tslint.json
jkhmtx: Perhaps better under a different scope? (See previous formatting comment)

Issue #1439, opened 2022-02-17
Inverisify InRequestScope does not work with toFactory binding;


Ayzrian:
## Expected Behavior
toFactory binding should be resolved as InRequestScope when requested;

## Current Behavior
It uses the context from the first resolvar of factory, since its scope is hardcoded to be singleton; Due to that, it creates a bug that in a new get request to inRequestScope binding, the factory will be using dependencies from previos request;

## Possible Solution
Expose a way to configure toFactory request scope;

## Steps to Reproduce (for bugs)

1. Create a toFactory binding;
2. Create a few InRequestScope bindings;
3. Request the in request scope bindings twice, and see that the second time uses instance from previous request;

## Context
I am trying to apply DI in azure funcitons, and each run azure function provides you a new logger, this new logger is registered in the child container, and it works for the first time, but when I request the same thing on the second call I see this in the logs 

Warning: Unexpected call to 'log' on the context object after function execution has completed. Please check for asynchronous calls that are not awaited or calls to 'done' made before function execution completes. Function name: article-feed-manual. Invocation Id: b0d77d73-e19f-4893-9968-4582126adee5. Learn more: https://go.microsoft.com/fwlink/?linkid=2097909

Though when I remove toFactory and creating instance with my hands everything works out :) 

## Your Environment

* Version used: 6
* Node 14
* MacOS

Ayzrian:
Resolved in 1440

PR #1440, opened 2022-02-17
docs: how to create factory with different scope

Ayzrian:
## Description
Allow to change toFactory scope;
## Related Issue
https://github.com/inversify/InversifyJS/issues/1439

## Motivation and Context

https://github.com/inversify/InversifyJS/issues/1439
## Types of changes

## Checklist:

### DIFF
--- a/src/syntax/binding_to_syntax.ts
+++ b/src/syntax/binding_to_syntax.ts
@@ -50,11 +50,11 @@ class BindingToSyntax<T> implements interfaces.BindingToSyntax<T> {
     return new BindingWhenOnSyntax<T>(this._binding);
   }
 
-  public toFactory<T2>(factory: interfaces.FactoryCreator<T2>): interfaces.BindingInWhenOnSyntax<T> {
+  public toFactory<T2>(factory: interfaces.FactoryCreator<T2>): interfaces.BindingWhenOnSyntax<T> {
     this._binding.type = BindingTypeEnum.Factory;
     this._binding.factory = factory;
     this._binding.scope = BindingScopeEnum.Singleton;
-    return new BindingInWhenOnSyntax<T>(this._binding);
+    return new BindingWhenOnSyntax<T>(this._binding);
   }
 
   public toFunction(func: T): interfaces.BindingWhenOnSyntax<T> {
@@ -84,11 +84,11 @@ class BindingToSyntax<T> implements interfaces.BindingToSyntax<T> {
     return new BindingWhenOnSyntax<T>(this._binding);
   }
 
-  public toProvider<T2>(provider: interfaces.ProviderCreator<T2>): interfaces.BindingInWhenOnSyntax<T> {
+  public toProvider<T2>(provider: interfaces.ProviderCreator<T2>): interfaces.BindingWhenOnSyntax<T> {
     this._binding.type = BindingTypeEnum.Provider;
     this._binding.provider = provider;
     this._binding.scope = BindingScopeEnum.Singleton;
-    return new BindingInWhenOnSyntax<T>(this._binding);
+    return new BindingWhenOnSyntax<T>(this._binding);
   }
 
   public toService(service: string | symbol | interfaces.Newable<T> | interfaces.Abstract<T>): void {
--- a/src/syntax/binding_to_syntax.ts
+++ b/src/syntax/binding_to_syntax.ts
@@ -84,11 +84,11 @@ class BindingToSyntax<T> implements interfaces.BindingToSyntax<T> {
     return new BindingWhenOnSyntax<T>(this._binding);
   }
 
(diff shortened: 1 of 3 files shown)


Ayzrian:
@LeonardSSH excuse me for bothering you, but any changes you could have a look at it?

leonardssh (approved):

leonardssh:
Unit Tests fails, take a look.

Ayzrian:
@LeonardSSH I have aligned tests, please have a look

Ayzrian:
Okay the pipelines are green now. Can we merge this PR, or I need to add something else as well?
@LeonardSSH

Ayzrian:
@LeonardSSH  can we merge this, please?

Ayzrian:
@LeonardSSH what do you think if I will change the approach to make it backward compatible? 

I will make `toFactory` return `BindInWhenOnSyntax`, so that by default the scope will be Singleton, while anyone could change scope if needed;

What do you think?

PodaruDragos:
@tonyhallett do you have any objection to this ?

Ayzrian:
@PodaruDragos  I have changed the implementation to be backward compatible; Now `toFactory` returns `InWhenOn` syntax;

tonyhallett:
@PodaruDragos
I am not the best person to ask about Inversify as I have not looked at the source or used it for quite some time now.
What I will say is that the latest commit is much more acceptable than the original.  What is missing from the latest commit is a corresponding change to the interface.

I would also say ( from my vague understanding of Inversiy ) is that any change to toFactory should equally apply to toProvider as it is just an async version of toFactory.

This may be a silly question but is it possible to use toDynamicValue ?

Ayzrian:
@tonyhallett  I am using `toDynamicValue` in my code, to bootstrap a logger each time a function is called, but the problem is that factories are not assembled in `request` scope so the first logger that Azure gave to your code will forever be stuck in the factory, that is incorrect since every call has unique logger from Azure;

So how it looks ;

1. I configure container;
2. When a function is called I create child container and bind logger to `toDynamicValue`.
3. Get an entry point entity, and call business logic;

But the next call factories that are injected in the entities are the same that were in the first call, because `toFactory` is always `Singleton`

Also I have fixed interfaces in b68ff8b

Ayzrian:
@tonyhallett I have did the same changes to `toProvider` in [97cf64e](https://github.com/inversify/InversifyJS/pull/1440/commits/97cf64e760a511946e0796776228c347f7111d55)

tonyhallett:
@Ayzrian
I meant can you use toDynamicValue for your factory.

Ayzrian:
@tonyhallett but this brings a problem that each function will need to pull the factories and the code that these factories envokes and create them in the child container; This can lead to the growing size of the bundle for each azure function and increased boot time, just because one of them is using a factory; 🤔  

Does it make any sense?

tonyhallett:
@Ayzrian 
I only mention it as internally they are treated the same.

https://github.com/inversify/InversifyJS/blob/8f3fd5750078b32b294761a3a365a3bb5218baea/src/resolution/resolver.ts#L85

Ayzrian:
@tonyhallett 

Hmm I got you; I didn't know that `toDynamicValue` receives a `context` as well; In that case `toDynamicValue` can be used as a workaround. Thank you!

IMHO, it would be more convenient and obvious to allow `toFactory` & `toProvider` to have `Request` scope, rather than applying workarounds with `toDynamicValue`

tonyhallett:
> IMHO, it would be more convenient and obvious to allow toFactory & toProvider to have Request scope, rather than applying workarounds with toDynamicValue

As I said, I am not well versed in Inversify as I used to be so I will leave it to @PodaruDragos and @LeonardSSH to decide.  Probably is more than fine to allow opening up the scope but perhaps a change to the documentation will be preferred.  Factory has had this fixed scope for many years, perhaps toDynamicValue is not a workaround but its pupose is to provide a value ( which can be a factory ) using context in the scope that you choose.  

The tyescript typing does suggest that the only way to do a factory is with toFactory.

That's all I have to say on the matteer !

Over to you @PodaruDragos @LeonardSSH

Ayzrian:
Thank you!
So @PodaruDragos , @LeonardSSH , what do you say? Should I update the documentation and we will merge this PR or should we close it?

PodaruDragos:
@Ayzrian I don't see any reason not to merge this, just please update the documentation.

@tonyhallett thanks for the input. I only tagged you since you were planning to update a lot of things with the v6 binding resolution refactor.

Ayzrian:
@PodaruDragos I have updated documentation, please have a look.

PodaruDragos:
cool, thanks. Will merge when the CI is green.

Issue #1409, opened 2021-12-17
toService not working for async services


AGrzes:
## Expected Behavior
At the moment it is possible to define service having asynchronous `@postConstruct()` method making it asynchronous and it would be great to have it work together with `toService` binding.

## Current Behavior
As toService is doing basically 
```
    BindingToSyntax.prototype.toService = function (service) {
        this.toDynamicValue(function (context) { return context.container.get(service); });
    };
```
It breaks as getAsync need to be used.
## Possible Solution
Add new api method
```
    BindingToSyntax.prototype.toAsyncService = function (service) {
        this.toDynamicValue(function (context) { return context.container.getAsync(service); });
    };
```
## Steps to Reproduce (for bugs)
```ts
import { Container, injectable, multiInject, postConstruct } from 'inversify'
import 'reflect-metadata'

@injectable()
class A {
  constructor(@multiInject('X') x) {
    console.log(x)
  }
}

@injectable()
class B {
  @postConstruct()
  async init() {}
}

const container = new Container()
container.bind(B).toSelf().inSingletonScope()
container.bind(A).toSelf()
container.bind('X').toService(B)
container.getAsync(A).then(console.log).catch(console.error)
```

`npx ts-node test.ts`

```
Error: You are attempting to construct 'class B {
    async init() { }
}' in a synchronous way
 but it has asynchronous dependencies.
    at Container._getButThrowIfAsync (/home/abramczykg/git/tuatara-finance-platform/tfp-local-development/modules/cli/node_modules/.pnpm/inversify@6.0.1/node_modules/inversify/src/container/container.ts:583:13)
    at Container.get (/home/abramczykg/git/tuatara-finance-platform/tfp-local-development/modules/cli/node_modules/.pnpm/inversify@6.0.1/node_modules/inversify/src/container/container.ts:325:17)
    at Binding.dynamicValue (/home/abramczykg/git/tuatara-finance-platform/tfp-local-development/modules/cli/node_modules/.pnpm/inversify@6.0.1/node_modules/inversify/src/syntax/binding_to_syntax.ts:96:38)
    at /home/abramczykg/git/tuatara-finance-platform/tfp-local-development/modules/cli/node_modules/.pnpm/inversify@6.0.1/node_modules/inversify/src/resolution/resolver.ts:51:86
[...]

PR #1441, opened 2022-02-18
Allow toService to be bound to async service 

AGrzes:
## Description

toService method:
```
 public toService(service: string | symbol | interfaces.Newable<T> | interfaces.Abstract<T>): void {
    this.toDynamicValue(
      (context) => context.container.get<T>(service)
    );
  }
```
Was using get method that throws exception when it would return promise. It caused prevented resolution of service using `getAsync` method.
`getAsync` could not be used instead because it always return promise even when resolution returns simple value (`async` keyword on function have that effect)
Removal of `async` keyword would change `getAsync` result - it would work for clients using `await` because one can await simple value but it would break for clients calling `then`/`catch` because simple values does not have them.
New function `getMaybeAsync` was introduced instead to Container interface returning promise when resolved to promise and simple value when resolved to simple value.

If introduction of new interface method is not desired then I see another way to approach problem - by passing if resolution is sync/async through Context.

## Related Issue

#1409

## Motivation and Context

toService binding were using Container.get method internally so it was throwing exception when the binding target was promise even if the resolution was also asynchroneous (getAsync)

## How Has This Been Tested?

Added new automated test case. 
All old test cases are passing. 

## Types of changes

## Checklist:

I will update changelog/documentation if general shape of PR is acceptable.

### DIFF
--- a/src/container/container.ts
+++ b/src/container/container.ts
@@ -331,6 +331,12 @@ class Container implements interfaces.Container {
     return this._get<T>(getArgs) as Promise<T> | T;
   }
 
+  public getMaybeAsync<T>(serviceIdentifier: interfaces.ServiceIdentifier<T>): Promise<T> | T {
+    const getArgs = this._getNotAllArgs(serviceIdentifier, false);
+
+    return this._get<T>(getArgs) as Promise<T> | T;
+  }
+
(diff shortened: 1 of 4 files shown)


notaphplover:
#1635 solved this issue

PR #1461, opened 2022-05-19
Add `allowLazyInSync` container option

paul-marechal:
This is useful when binding promise objects while still using the
synchronous APIs.

## Description

Add a new `allowLazyInSync` option to `interfaces.ContainerOptions`.

It is `false` by default.

When `true`, it will prevent throwing when getting promise objects from the bindings.

## Related Issue

n/a

## Motivation and Context

We used an old version of Inversify (5.1.1) and wanted to upgrade to 6.0.1.

We used to bind some promise objects as `.toConstantValue(promise...)` and have the components deal with it after construction. This now seem to completely break with the new Inversify logic.

We don't have time to completely refactor our code to avoid the error from being thrown, so turning the check off is our best bet for now.

## How Has This Been Tested?

Added a new unit test for this option.

## Types of changes

## Checklist:

### DIFF
--- a/src/container/container.ts
+++ b/src/container/container.ts
@@ -65,6 +65,14 @@ class Container implements interfaces.Container {
       throw new Error(`${ERROR_MSGS.CONTAINER_OPTIONS_MUST_BE_AN_OBJECT}`);
     }
 
+    if (options.allowLazyInSync === undefined) {
+      options.allowLazyInSync = false;
+    } else if (
+      typeof options.allowLazyInSync !== "boolean"
+    ) {
+      throw new Error(ERROR_MSGS.CONTAINER_OPTIONS_INVALID_ALLOW_LAZY_IN_SYNC);
+    }
+
     if (options.defaultScope === undefined) {
       options.defaultScope = BindingScopeEnum.Transient;
     } else if (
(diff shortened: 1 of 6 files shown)


PodaruDragos:
I think this is fine.
@dcavanagh @jakehamtexas what do you guys think ?

@paul-marechal can you please make sure the CI get's to green ?

src/container/container.ts
```diff
@@ -579,7 +588,7 @@ class Container implements interfaces.Container {
   ): (T | T[]) {
     const result = this._get<T>(getArgs);
 
-    if (isPromiseOrContainsPromise<T>(result)) {
+    if (!this.options.allowLazyInSync && isPromiseOrContainsPromise<T>(result)) {
       throw new Error(ERROR_MSGS.LAZY_IN_SYNC(getArgs.serviceIdentifier));
```
paul-marechal, line 592: `getArgs.serviceIdentifier` can be a symbol here, and it fails to get converted to string in `LAZY_IN_SYNC`.

I added a quick fix in `error_msgs.ts` to not fail in such a case. Not sure if the impl is good:

```ts
const TO_STRING = (value: any) => typeof value?.toString === 'function' ? value.toString() : `${value}`;
```

Surprisingly, doing the following fails:

```ts
const symbol = Symbol('test');
console.log(`${symbol}`);
```

But calling `.toString()` manually on the symbol works... Gotta love working with JavaScript :)

paul-marechal:
@PodaruDragos CI should be fixed.

leonardssh (approved):
LGTM!

paul-marechal:
Closing because I realized it might be weird to try and bend Inversify in way it isn't meant to be.

I don't know what you guys think about the constraint with never binding promises/thenables but it sounds like it makes sense to encourage binding actual implementations and prevent raw promises from being passed around?

PR #1506, opened 2023-04-13
Add .toDynamicValueWithDeps into BindingToSyntax

SamakaCD:
`toDynamicValueWithDeps` method binds an abstraction to a dynamic value with required dependencies from the container in a declarative way.

## Description

I've addded `toDynamicValueWithDeps` method into BindingToSyntax class which basically uses `toDynamicValue` method and context it provides under the hood of the new method.

## Related Issue

None.

## Motivation and Context

Let's say we have the following classes:
```ts
abstract class AbstractShuriken {}

abstract class AbstractKatana {}

@injectable()
class Shuriken implements AbstractShuriken {}

@injectable()
class Katana implements AbstractKatana {}

class Ninja {
  public constructor(public shuriken: AbstractShuriken, public katana: AbstractKatana) {}
}
```

I would like to create a binding for `Ninja` class which depends on abstract shuriken and katana. Without the new method we would do something like this:
```ts
container.bind(Ninja).toDynamicValue((context) => {
  const shuriken = context.container.get(AbstractShuriken)
  const katana = context.container.get(AbstractKatana)
  return new Ninja(shuriken, katana)
})
```

The code might be even longer, especially when it's required to inject more dependencies. To make binding notation shorter we could introduce a method which accepts a list of dependencies and a factory which `toDynamicValueWithDeps` method actually does:
```ts
container.bind(Ninja).toDynamicValueWithDeps(
  [AbstractShuriken, AbstractKatana] as const,
  ([shuriken, katana]) => new Ninja(shuriken, katana)
)
```
Looking forward, it should be possible to remove `as const` for dependencies using [const type parameters](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters) when the library will use TypeScript 5.

## How Has This Been Tested?

I've added `should be able to resolve all dependencies using toDynamicValueWithDeps` test case into `container.test.ts`.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `.toDynamicValueWithDeps` to create a dynamic value with declaratively listed dependencies
 
 ### Changed
 
--- a/src/interfaces/interfaces.ts
+++ b/src/interfaces/interfaces.ts
@@ -340,6 +340,13 @@ namespace interfaces {
     toSelf(): BindingInWhenOnSyntax<T>;
     toConstantValue(value: T): BindingWhenOnSyntax<T>;
     toDynamicValue(func: DynamicValue<T>): BindingInWhenOnSyntax<T>;
+    toDynamicValueWithDeps<Deps extends readonly unknown[]>(
+      dependencies: Deps,
+      func: (
+        dependencies: ResolvedDeps<Deps>,
+        context: interfaces.Context
+      ) => T
+    ): interfaces.BindingInWhenOnSyntax<T>;
     toConstructor<T2>(constructor: Newable<T2>): BindingWhenOnSyntax<T>;
     toFactory<T2, T3 extends unknown[] = unknown[], T4 extends unknown[] = unknown[]>(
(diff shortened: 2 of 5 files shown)


PodaruDragos:
This add a new method to the binding, so I am a bit reticent about this. code looks good though.
Might need more opinions on this.

PS: you could always just build this on your own project, TBH

SamakaCD:
> you could always just build this on your own project, TBH

And that's right 🙂. Initially, I tried to extend `ContainerModule` class to replace the original binding syntax. But the [result](https://gist.github.com/SamakaCD/19b4789f290be7c839f322ded73ef455) was somewhat cumbersome 😅

Anyway, it is up to you whether to include such a binding extension.

SamakaCD:
Improved generic constraints for `toDynamicValueWithDeps`. Also, updated the "Motivation and Context" section to highlight that it's not required to decorate binding class / service identifier with `@injectable()`.

notaphplover:
Closing the issue as the codebase is already migrated to the monorepo.

Adding some notes for future implementations:

1. Let's give the `Factory` binding this feature instead.
2. Let's not rely on `container.get` under any circumstances. The planner must be updated instead to support factory dependencies as it handles class dependencies. This way we can provide optional, multiple and async dependencies. This way we don't mix the planning and the resolution phase.

PR #1527, opened 2023-09-13
ci: run ci against node `18.x, 20.x`

### LINKED ISSUE #1526: Tests for #549 are flaky (node 10 / es6)
The tests for #549 are flaky on CI when running against `10.x` and `src/tsconfig-es6.json`

- https://github.com/inversify/InversifyJS/actions/runs/6163690582/job/16727801192
- https://github.com/inversify/InversifyJS/actions/runs/6163820009/job/16728207564

```
  1 failing

  1) Issue 549
       Should throw if circular dependencies found with dynamics:
     Error: Expected `willThrow` to throw:
- It looks like there is a circular dependency in one of the 'toDynamicValue' bindings. Please investigate bindings with service identifier 'Symbol(ADynamicValue)'.
or
- It looks like there is a circular dependency in one of the 'toDynamicValue' bindings. Please investigate bindings with service identifier 'Symbol(BDynamicValue)'.
but got
- Cannot read property 'rootRequest' of undefined
      at Context.it (test/bugs/issue_549.test.ts:65:15)
      at process.topLevelDomainCallback (domain.js:126:23)
```

Jameskmonger:
fixes #1526

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
         ts-project: [src/tsconfig.json, src/tsconfig-es6.json]
     env:
       TS_NODE_PROJECT: ${{ matrix.ts-project }}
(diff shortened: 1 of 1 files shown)


PodaruDragos (approved):
can we just leave 18 and 20 ? 
16 is EOL. so no point in including 16

Jameskmonger:
@PodaruDragos happy to do so if you think that's best, I left it in because of how recently EOL for 16.x was.

Issue #1455, opened 2022-05-09
TYPES must be in own file!?


andrecadete:
Hi, not really a bug report, just to see if you are aware since this is not obvious in the docs. As title states.

In fact I cannot get this to work unless the TYPES are explicitly declared and exported in their own file. For example, if the TYPES are exported and declared in the same file as the Container extension class, we will get something: "TypeError: Cannot read property 'DynamoDB' of undefined", where `undefined` are our TYPES.

Please let me know if this is expected behavior and if so I'll open a PR to improve the docs' clarity in that sense.

Thanks

AkshayJangid:
Hey, I'll be more then happy to make a PR for this. Let me know if i can help here, but lets add this ASAP.

shellscape:
@Jameskmonger given your comments here https://github.com/inversify/InversifyJS/pull/1546#issuecomment-1789488948, this is probably worthy of a look given its age, no?

notaphplover:
I don't see how this is related to inversify. After having a look at #1179, `Types` are already imported from their own file. I'm closing the issue now. If you want feel free to provide a minimum reproduction code in a [codesandbox](https://codesandbox.io) or a reproduction repo.

shellscape:
A year and 5 days!

PR #1546, opened 2023-10-31
Update README.md

AkshayJangid:
by going through the documentation , it is no where hinted that we need to place Types declaration in separate file. Thus explicitly mentioning it in readme.

Updated Readme file to explicitly mention about adding Types declaration in separate file. 

## Description

By going through the documentation it is not very clear about adding this in separata file. This leads to error 
Cannot read properties of undefined on the leaf elements. 

## Motivation and Context

I was stuck because of this for 4 hours. 

## How Has This Been Tested?

Not required. 

## Types of changes

- [x ] Updated docs / Refactor code / Added a tests case (non-breaking change)

## Checklist:

- [ ] My code follows the code style of this project.
- [x ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] All new and existing tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,9 @@ export interface ThrowableWeapon {
 ```
 
 InversifyJS needs to use the type as identifiers at runtime. We use symbols as identifiers but you can also use classes and or string literals.
-PLEASE MAKE SURE TO PLACE THIS TYPES DECLARATION IN A SEPARATE FILE. 
+
+PLEASE MAKE SURE TO PLACE THIS TYPES DECLARATION IN A SEPARATE FILE. (see bug #1455)
+
 ```ts
 // file types.ts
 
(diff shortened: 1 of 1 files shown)


Jameskmonger:
Hi @AkshayJangid , thanks for the submission - I don't think this is intended behaviour but it would be good to document it while it's still present. I'm just catching up on #1455. It might be good to link to this issue in the README, as my understanding is that this is a bug and should be fixed.

AkshayJangid:
Hey @Jameskmonger , the problem is at least 4 people i know were stuck because of this for a day. Hence just trying to add it temporarily.

Jameskmonger:
@AkshayJangid  I completely agree. Please adjust the PR to have:

```
bug: PLEASE MAKE SURE TO PLACE THIS TYPES DECLARATION IN A SEPARATE FILE. (see bug #1455)
```

then I will merge it immediately. Ty

Jameskmonger:
Looks like I can edit your PR 😄  I will merge now, ty for contrib @AkshayJangid

PR #1572, opened 2024-06-11
Update post_construct.md

alexander-joinlane:
Seems to be an inaccuracy in the Post Construct Decorator description (implementation seems correct)

## Description

* `use()` returned a string, no visible side-effect → updated to `console.log()`
* `testMethod()` was not discernibly different from `use()` → updated to `"Used ..."` and `"Tested ..."`
* `use()` was not invoked → updated to invoke so that the log output would make more sense

I did this via Github interface, so conventional commits didn't seem to be available

## Related Issue

N/A

## Motivation and Context

Clarity of use and intent.

## How Has This Been Tested?

N/A

## Types of changes

N/A

## Checklist:

N/A

### DIFF
--- a/wiki/post_construct.md
+++ b/wiki/post_construct.md
@@ -13,18 +13,18 @@ The method can be synchronous or asynchronous.
 
 
 ```ts
-interface Katana {
-    use: () => void;
+interface Weapon {
+  use: () => void;
 }
 
 @injectable()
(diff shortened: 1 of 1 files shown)


notaphplover:
Closing as these docs are legacy and already replaced by new ones in the monorepo

Issue #1616, opened 2024-11-11
Webpack uses incorrect entry point as of v6.1.1


msujew:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I am a maintainer of the Theia IDE framework, and was made aware of this issue in https://github.com/eclipse-theia/theia/discussions/14430. After v6.1.0, webpack no longer correctly find the cjs entry point and instead decides to use the ESM entry point. This eventually results in an obscure webpack error, complaining that `@inversifyjs/core` has no exports at all.

### Steps to reproduce

1. Update inversify to 6.1.1 in [here](https://github.com/eclipse-theia/theia/blob/bc4acd43e5a33dfe64c9bc3fb7ccb6ae9189c76f/packages/core/package.json#L54).
2. Run `yarn && yarn browser build`.
3. Webpack will eventually report the error reported in https://github.com/eclipse-theia/theia/discussions/14430.

### Expected behavior

Webpack continues using the cjs entry point as was the case in 6.0.3

### Possible solution

In my local environment, I've added the following lines of export configuration (which already exist in `@inversifyjs/core`). This seems to fix the issue again:

```json
"exports": {
  ".": {
    "import": "./es/inversify.js",
    "require": "./lib/inversify.js"
  }
}
```

### Package version

6.1.1

### Node.js version

18.20.1

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [X] Linux

### Stack trace

```
WARNING in ../../node_modules/inversify/es/planning/reflection_utils.js 10:22-32
export 'getTargets' (imported as 'getTargets') was not found in '@inversifyjs/core' (module has no exports)
@ ../../node_modules/inversify/es/planning/planner.js 11:0-100 114:31-46 116:47-74 118:69-84
@ ../../node_modules/inversify/es/container/container.js 62:0-85 122:19-39 124:32-52 317:28-45 637:26-30
@ ../../node_modules/inversify/es/inversify.js 4:0-50 4:0-50
@ ./src-gen/backend/server.js 12:22-42
@ ./src-gen/backend/main.js 77:21-40
ERROR in ../../node_modules/@inversifyjs/common/lib/esm/index.js 1:0-80
Module not found: Error: Can't resolve './services/models/LazyServiceIdentifier' in '<project_path>\node_modules@inversifyjs\common\lib\esm'
Did you mean 'LazyServiceIdentifier.js'?
[...]

PR #1617, opened 2024-11-11
Add exports field to fix webpack errors

msujew:
## Description

Adds a `exports` field to direct bundlers (mainly webpack) to the correct entry point. Having that entry in there is good practice anyway, and the other `@inversifyjs/*` packages already have an `exports` field.

## Related Issue

https://github.com/inversify/InversifyJS/issues/1616

## Motivation and Context

I assume other people will have a similar issue when it comes to using webpack and CJS (i.e. `require`) of inversify.

## How Has This Been Tested?

Applying this change to my local environment has fixed the issue reported in https://github.com/inversify/InversifyJS/issues/1616.

## Types of changes

## Checklist:


sdirix:
I tested the change locally on my system. It successfully fixes the issue in which webpack consumed `inversify`'s esm build instead of commonjs although inversify was `require`d.

package.json
```diff
@@ -48,6 +48,13 @@
   "license": "MIT",
   "main": "lib/inversify.js",
   "module": "es/inversify.js",
+  "exports": {
+    ".": {
+      "types": "./lib/inversify.d.ts",
```
notaphplover:
```suggestion
```
If we go this way, tsc will default to cjs types even when ESM imports are used. Is this line critical to solve the issue?

msujew: No, I just wasn't entirely sure how tsc will behave. Thank you for the explanation! I've removed the line again :)

notaphplover (approved):
Thank you so much :heart:

